PR #5, opened 2026-06-29
feat(ethclspecs): Gloas EIP-8282 + re-pin consensus-specs to v1.7.0-alpha.11

IvanAnishchuk:
Stacked on #4, which is the base of this branch. Until #4 merges the diff here
shows both commits (the libcrypto build fix from #4 plus the Gloas work); it
drops to just the Gloas commit once #4 lands.

This re-pins EthCLSpecs from consensus-specs v1.7.0-alpha.10 to v1.7.0-alpha.11
and brings Gloas up to the alpha.11 shape. I want it in mainly as the base for the
Heze (EIP-7805 FOCIL) work I'm doing next, which stacks on top of this.

One caveat up front: I'm not sure whether anyone else has already started the
alpha.11 Gloas update. If someone has a better implementation going I'm happy to
rebase onto it or step aside, just say so.

## EIP-8282 (builder deposits / exits)

New `BuilderDepositRequest` / `BuilderExitRequest` containers, two more lists on
`ExecutionRequests` (`builder_deposits`, `builder_exits`), a `proposer_index`
field on `BuilderPendingPayment`, and the matching constants.

## Keep-green for the alpha.11 spec diff

The conformance run surfaced two changes the spec text didn't telegraph:
- `process_execution_payload_bid` now takes the signed bid directly and reads slot
  and parent root from state, with a payload-builder-version check.
- `process_deposit_request` and `process_voluntary_exit` fall back to the inherited
  handlers; builder onboarding and exit moved to the dedicated EL-triggered requests.
- `process_proposer_slashing` only clears a builder payment it actually owns, and
  the fork bid root uses the 5-list `ExecutionRequests`.

## Deferred

EIP-8282 builder-deposit/exit operation *processing* is still xfail. Follow-up,
not wired here yet.

## Conformance

gloas minimal: 6542 passed, 0 failed. fulu: 152p/44xf (unchanged). gloas mainnet:
1003 passed, 0 failed (161 xfail).

### DIFF
--- a/packages/EthCLLib/lakefile.toml
+++ b/packages/EthCLLib/lakefile.toml
@@ -6,7 +6,9 @@ defaultTargets = ["EthCLLib"]
 # `libcrypto` flag its `native_decide` self-tests need transitively through
 # SizzLean's FFI SHA-256. The blst / c-kzg archives propagate automatically
 # as `extern_lib`s, only the system shared lib needs re-stating here.
-moreLinkArgs = ["-L/usr/lib/x86_64-linux-gnu", "-l:libcrypto.so.3"]
+# Two `-L` paths cover the common Linux libdir layouts (Debian multiarch and
+# `/usr/lib64` on Gentoo / Fedora / RHEL); the linker ignores whichever is absent.
+moreLinkArgs = ["-L/usr/lib/x86_64-linux-gnu", "-L/usr/lib64", "-l:libcrypto.so.3"]
 moreLeancArgs = ["-march=native"]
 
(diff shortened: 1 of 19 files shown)


leolara:
@IvanAnishchuk 

 maxBuilderDepositRequestsPerPayload and maxBuilderExitRequestsPerPayload are preset and should be added like presets.

leolara:
> @IvanAnishchuk
> 
> maxBuilderDepositRequestsPerPayload and maxBuilderExitRequestsPerPayload are preset and should be added like presets.

We will fix this in a followup PR

PR #6, opened 2026-06-30
feat(ethclspecs): add Heze (EIP-7805 FOCIL) for v1.7.0-alpha.11

IvanAnishchuk:
Adds the Heze fork layer (EIP-7805 FOCIL) at consensus-specs v1.7.0-alpha.11. Heze at this
pin is a thin diff over Gloas: ethereum/consensus-specs#5371 reverted the inclusion-list
bitlist back out of the bid, so the consensus delta is the `InclusionList` container family
plus the FOCIL fork-choice machinery.

## Conformance

heze minimal: 6770 passed, 3198 skipped, 0 failed, 0 xfail. heze mainnet: 997 passed, 130
skipped, 0 failed, 0 xfail. Every heze `fork_choice` case passes (84 minimal + 53 mainnet),
including through the two FOCIL overrides with behavioral deltas on the vector path
(`should_extend_payload`, `on_execution_payload_envelope`; the re-declared store constructor
also runs everywhere, its delta is seeding the two new fields empty). fulu and gloas are
unchanged. Deliberately excluded categories stay visible: the
`ssz_static` types Heze doesn't model and the `genesis` runner report `skip`, and the
harness now names its one never-collected runner (`fast_confirmation`) in the allowlist
comment instead of leaving the omission silent.

## What's in the layer

- The Gloas spine re-instantiated over Heze's byte-identical twin types (EpochProcessing /
  Operations / Withdrawals / Transition / ForkChoice, each inheriting its Gloas counterpart).
  EIP-7805 adds no new state-transition step at alpha.11.
- The two new containers, `InclusionList` / `SignedInclusionList`, the passthrough
  `upgradeToHeze`, and the beacon-chain helpers `get_inclusion_list_committee` /
  `is_valid_inclusion_list_signature`.
- The FOCIL fork-choice layer: the `InclusionListStore`, `process_inclusion_list` /
  `get_inclusion_list_transactions`, `on_inclusion_list`, the satisfaction record/gate pair,
  `get_inclusion_list_due_ms`, and the two overrides above.

Alpha.11 ships `ssz_static` vectors for the two new containers (full suites, passing) but no
vector driving FOCIL behavior: no `on_inclusion_list` fork-choice step, no operation, nothing
feeding the inclusion-list store. So everything past the two overrides runs on no behavioral
conformance input yet. I mirrored that layer branch-for-branch from the pinned spec text and
pinned it with build-enforced `#guard`s / `native_decide` examples (expected values worked out
by hand), so a later edit can't regress it silently. `is_inclusion_list_satisfied` is an
Engine-API verdict with no EL in this harness; it enters through a framework-level
`[ExecutionEngine]` seam (a generic class in EthCLLib, its own commit) whose default
instance answers `true` (the treatment Gloas gives `verify_and_notify_new_payload` /
`is_data_available`), and a refuting-oracle pin drives the `false` branch no vector can
reach. The vectorless layer's deliberate spec-text divergences are recorded in
`packages/EthCLSpecs/docs/DISCREPANCIES.md`.

## One supporting fix (touches Fulu and Gloas)

The fork-choice step interpreters silently passed unmodeled steps through a `| _ => pure ()`
catch-all, so a `checks.get_proposer_head` entry scored a vacuous pass. All three
interpreters now match `FcStep` exhaustively: unmodeled steps surface as `todo` (an xfail)
and a newly added constructor is a build error. The two step decoders (harness and server)
report an unrecognized step or check key as `unsupported` instead of dropping it, which
pre-wires a future `on_inclusion_list` vector. No alpha.11 vector hits any of these paths,
so conformance is unchanged. The `pyspec_server` CLI got the same treatment: an unknown
fork or preset name errors with the expected-names list instead of silently serving Fulu.
Two housekeeping commits ride along: a `PYTEST_JOBS` override for the full-sweep recipes
(default unchanged) and a docs sweep of the pin citations the alpha.11 re-pin left stale.

## Known limitation

`fcInterpretHeze` is a near-verbatim copy of `fcInterpretGloas`, differing only in the
fork-namespaced decode types. The state-transition spine inherits across forks; the step
interpreter has no equivalent seam yet, so each fork clones the whole match and keeps it in
lockstep by hand. The natural fix is a parametric interpreter over a fork's decode bundle; I
queued that as a follow-up rather than folding it into a fork-addition PR, and the same goes
for migrating Gloas's two inline engine constants onto the new `[ExecutionEngine]` seam.

<details>
<summary><b>Deferred follow-ups</b> (tracked on my side; none block this PR). Tick anything you'd like filed as an upstream issue (or pulled into this PR) and I'll take it from there.</summary>

- [ ] **`fast_confirmation` runner**: the corpus ships a whole fork-choice format we don't collect yet.
  <details><summary>details</summary>

  The alpha.11 corpus carries `fast_confirmation` (the confirmation-rule format) for
  fulu/gloas/heze, minimal-only, ~1530 yaml files across 9 handlers. It is deliberately
  uncollected and now named in the
  [`IN_SCOPE_RUNNERS` allowlist comment](https://github.com/IvanAnishchuk/etheorem/blob/feat/heze-focil-alpha11/packages/EthCLSpecs/PySpecTests/harness.py).
  Implementing it is real fork-choice work: a steps script plus the confirmation-rule store
  fields (`previous_slot_head`, observed-justified / greatest-unrealized checkpoints), and
  upstream is still growing the format (~870 new lines of phase0 tests post-alpha.11).
  </details>
- [ ] **Parametric fork-choice interpreter**: the clone family from the Known-limitation section.
  <details><summary>details</summary>

  Beyond `fcInterpretHeze` ≈ `fcInterpretGloas`: `runStatic` and `decodeOp` are fully
  fork-agnostic yet exist as three identical copies per
  [fork `Interface.lean`](https://github.com/IvanAnishchuk/etheorem/blob/feat/heze-focil-alpha11/packages/EthCLSpecs/EthCLSpecs/Heze/Interface.lean),
  and the ~45-arm `ssz_static` name dispatch is restated per fork, where a missed arm doesn't
  fail, it silently reports skip. A parametric interpreter over a fork's decode bundle plus a
  delta-based dispatch table fixes the class; kept out of here so the diff stays on the Heze layer.
  </details>
- [ ] **Gloas engine predicates onto `[ExecutionEngine]`**: finish what the seam commit starts.
  <details><summary>details</summary>

  [`EthCLLib/Spec/Engine.lean`](https://github.com/IvanAnishchuk/etheorem/blob/feat/heze-focil-alpha11/packages/EthCLLib/EthCLLib/Spec/Engine.lean)
  now owns the generic engine seam and Heze's `is_inclusion_list_satisfied` reads it;
  `verify_and_notify_new_payload` / `is_data_available` in
  [`Gloas/ForkChoice.lean`](https://github.com/IvanAnishchuk/etheorem/blob/feat/heze-focil-alpha11/packages/EthCLSpecs/EthCLSpecs/Gloas/ForkChoice.lean)
  are still inline constants. Migrating them gives one EL trust boundary and makes
  refuting-instance pins possible for both.
  </details>
- [ ] **Harness check-emitter registry**: make the known-key drift impossible, not just commented.
  <details><summary>details</summary>

  `KNOWN_CHECK_KEYS` / `KNOWN_CHECK_SUBKEYS` in
  [`harness.py`](https://github.com/IvanAnishchuk/etheorem/blob/feat/heze-focil-alpha11/packages/EthCLSpecs/PySpecTests/harness.py)
  hand-mirror the emitter chain in `_build_fork_choice_request`; a handler added without the
  set-update silently demotes a fully-checked case to xfail. One key→(emitter, subkeys) table
  driving both ends the failure mode.
  </details>
- [ ] **`checkHead` missing-block fallback**: replace a vacuous-pass-shaped default with a thrown error.
  <details><summary>details</summary>

  All three interpreters read the head block's slot as
  `match FcMap.lookup store.blocks head.root with | some b => b.slot.toNat | none => 0`. A store
  corrupt enough to make `getHead` return an unstored root (unreachable today; pyspec would
  KeyError loudly) could then pass a `head <root> 0` check vacuously. Cross-fork one-arm change.
  </details>
- [ ] **`should_override_forkchoice_update` bucket**: xfail today, arguably skip.
  <details><summary>details</summary>

  It routes `unsupported` → `todo` (the work-queue xfail bucket), but nothing intends to model
  it (it needs an EL head/safe-block signal), so by the
  [`Errors.lean`](https://github.com/IvanAnishchuk/etheorem/blob/feat/heze-focil-alpha11/packages/EthCLLib/EthCLLib/Spec/Errors.lean)
  taxonomy it may belong in `outOfScope` (skip). Vector-dead either way at alpha.11; a one-arm
  decision plus a comment.
  </details>
- [ ] **Preset watch: `INCLUSION_LIST_COMMITTEE_SIZE`**: nothing to do until presets diverge.
  <details><summary>details</summary>

  A preset-file value modeled as a universal-tier `abbrev` because both presets say `2**4` at
  this pin (and still on master). The comment in
  [`Fulu/Constants.lean`](https://github.com/IvanAnishchuk/etheorem/blob/feat/heze-focil-alpha11/packages/EthCLSpecs/EthCLSpecs/Fulu/Constants.lean)
  records the promotion path (a `Preset` field like `ptcSize`) for the day a re-pin diverges them.
  </details>
</details>

### DIFF
--- a/packages/EthCLSpecs/EthCLSpecs/Heze/Block.lean
+++ b/packages/EthCLSpecs/EthCLSpecs/Heze/Block.lean
@@ -5,9 +5,9 @@ import EthCLSpecs.Heze.Containers.InclusionList
 
 `BeaconBlockBody` / `BeaconBlock` / `SignedBeaconBlock` are Gloas's, unchanged (EIP-7805
 touches none of them), so all three are `inherit`ed verbatim, including the signed wrapper.
-The `Containers.InclusionList` import is the load-order entry (it pulls `Heze.Inherited` +
-the `fork Heze from Gloas` lineage that `inherit` resolves against); it is load-bearing,
-do not drop it.
+The `Containers.InclusionList` import looks unused but is required: it transitively loads
+`Heze.Inherited` and the `fork Heze from Gloas` declaration that `inherit` resolves
+against. Do not remove it.
 -/
 
 set_option autoImplicit false
--- a/packages/EthCLSpecs/EthCLSpecs/Heze/Block.lean
+++ b/packages/EthCLSpecs/EthCLSpecs/Heze/Block.lean
@@ -1,0 +1,24 @@
+import EthCLSpecs.Heze.Containers.InclusionList
+
+/-!
+# `EthCLSpecs.Heze.Block`: the Heze block containers (inherited from Gloas)
+
+`BeaconBlockBody` / `BeaconBlock` / `SignedBeaconBlock` are Gloas's, unchanged (EIP-7805
+touches none of them), so all three are `inherit`ed verbatim, including the signed wrapper.
+The `Containers.InclusionList` import is the load-order entry (it pulls `Heze.Inherited` +
+the `fork Heze from Gloas` lineage that `inherit` resolves against); it is load-bearing,
+do not drop it.
+-/
+
+set_option autoImplicit false
+
+open EthCLLib.Spec
+open EthCLSpecs.Fulu
+
+namespace EthCLSpecs.Heze
+
+inherit BeaconBlockBody
+inherit BeaconBlock
+inherit SignedBeaconBlock
+
+end EthCLSpecs.Heze
--- a/packages/EthCLSpecs/EthCLSpecs/Heze/Interface.lean
+++ b/packages/EthCLSpecs/EthCLSpecs/Heze/Interface.lean
@@ -8,15 +8,18 @@ import EthCLSpecs.Heze.ForkChoice
 /-!
 # `EthCLSpecs.Heze.Interface`: the Heze fork-interface instance
 
-Heze's implementation of `ForkInterface`. At v1.7.0-alpha.11 EIP-7805 (FOCIL) adds no state
-transition and no *vector-tested* fork-choice change: its overrides run on the existing fork_choice
-vectors but stay Gloas-equivalent (empty inclusion-list store), and the FOCIL-specific behavior
-ships no vector, so every dynamic runner reuses the Gloas spine re-instantiated over Heze types
-(`Heze.EpochProcessing` / `Operations` / `Withdrawals` /
-`Transition` / `ForkChoice`). `runUpgrade` is the Gloas→Heze `fork` format (a pure field
-copy, no onboarding); `runTransition` is the Gloas→Heze boundary; `runForkChoice` runs the
-Heze ePBS store. `runGenesis` and any `ssz_static` type Heze does not model are
-`outOfScope` (reported `skip` rather than counted as xfail work), matching Fulu / Gloas. Pinned to
+Heze's implementation of `ForkInterface`, the entry points the pyspec runner drives
+(`SPEC_AUTHORING_MODEL.md` §11). At v1.7.0-alpha.11 EIP-7805 (FOCIL) adds no state transition
+and no vector-tested fork-choice change. Its fork-choice overrides do run on the existing
+fork_choice vectors, but only on the path where the inclusion-list store stays empty, which
+behaves exactly like Gloas; the FOCIL-specific behavior ships no vector. So every dynamic
+runner reuses the Gloas spine re-instantiated over Heze types (`Heze.EpochProcessing` /
+`Operations` / `Withdrawals` / `Transition` / `ForkChoice`).
+
+The Heze-specific entries: `runUpgrade` is the Gloas→Heze `fork` format (a pure field copy,
+see `upgradeToHeze`); `runTransition` is the Gloas→Heze boundary; `runForkChoice` runs the
+Heze ePBS store. `runGenesis` and any `ssz_static` type Heze does not model report
+`outOfScope`, a deliberate skip rather than xfail work, matching Fulu / Gloas. Pinned to
 v1.7.0-alpha.11.
 -/
 
@@ -45,8 +48,7 @@ private def stateRootImpl (P : Preset) (bytes : ByteArray) :
 
 /-- `runUpgrade` (the `fork` format, Gloas→Heze): decode the Gloas pre-state at preset `P`,
 apply `upgradeToHeze` with the config's `HEZE_FORK_VERSION`, return the Heze post root. A
-pure field copy; no onboarding / PTC recompute (builders are already onboarded in the Gloas
-state). -/
+pure field copy; `upgradeToHeze`'s docstring carries the why. -/
 private def runUpgradeImpl (P : Preset) (forkVersion : Version) (preBytes : ByteArray) :
     Except (RunError StateTransitionError) ByteArray :=
   letI : Preset := P
@@ -187,8 +189,8 @@ private def runBlocksImpl (P : Preset) (C : Config) (preBytes : ByteArray)
 
 /-- `runTransition` (the `transition` format, Gloas→Heze): fold the pre-fork blocks under
 Gloas `state_transition`, advance the Gloas state to the fork-epoch boundary, apply
-`upgradeToHeze` (pure copy, no onboarding), then fold the post-fork blocks under the Heze
-spine. Unqualified spine names resolve to the Heze copies; the pre-fork Gloas calls are
+`upgradeToHeze` (a pure copy, see its docstring), then fold the post-fork blocks under the
+Heze spine. Unqualified spine names resolve to the Heze copies; the pre-fork Gloas calls are
 qualified. -/
 private def runTransitionImpl (P : Preset) (C : Config) (forkVersion : Version)
     (preBytes : ByteArray) (blocks : Array ByteArray) (cmeta : CaseMeta) :
--- a/packages/EthCLSpecs/EthCLSpecs/Heze/Interface.lean
+++ b/packages/EthCLSpecs/EthCLSpecs/Heze/Interface.lean
@@ -1,0 +1,402 @@
+import EthCLSpecs.Heze.Upgrade
+import EthCLSpecs.Heze.EpochProcessing
+import EthCLSpecs.Heze.Operations
+import EthCLSpecs.Heze.Withdrawals
+import EthCLSpecs.Heze.Transition
+import EthCLSpecs.Heze.ForkChoice
+
+/-!
+# `EthCLSpecs.Heze.Interface`: the Heze fork-interface instance
+
+Heze's implementation of `ForkInterface`. At v1.7.0-alpha.11 EIP-7805 (FOCIL) adds no state
+transition and no *vector-tested* fork-choice change: its overrides run on the existing fork_choice
+vectors but stay Gloas-equivalent (empty inclusion-list store), and the FOCIL-specific behavior
+ships no vector, so every dynamic runner reuses the Gloas spine re-instantiated over Heze types
+(`Heze.EpochProcessing` / `Operations` / `Withdrawals` /
+`Transition` / `ForkChoice`). `runUpgrade` is the Gloas→Heze `fork` format (a pure field
+copy, no onboarding); `runTransition` is the Gloas→Heze boundary; `runForkChoice` runs the
+Heze ePBS store. `runGenesis` and any `ssz_static` type Heze does not model are
+`outOfScope` (reported `skip` rather than counted as xfail work), matching Fulu / Gloas. Pinned to
+v1.7.0-alpha.11.
+-/
+
+set_option autoImplicit false
+
+open EthCLLib.Spec
+open EthCLLib.PySpecTests
+open EthCLSpecs.Fulu
+open SizzLean
+open SizzLean.Cache
+open SizzLean.Hasher
+
+namespace EthCLSpecs.Heze.Interface
+
+/-- Pinned upstream release; Heze tracks the same tag as Fulu / Gloas. -/
+def pyspecPinnedVersion : String := "v1.7.0-alpha.11"
+
+/-- `stateRoot`: decode a Heze `BeaconState` at preset `P` and take its root. -/
+private def stateRootImpl (P : Preset) (bytes : ByteArray) :
+    Except (RunError StateTransitionError) ByteArray :=
+  letI : Preset := P
+  letI : HasherTag := fastHasherTag
+  match SSZ.deserialize (T := @Heze.BeaconState P) bytes with
+  | .ok v    => .ok (htr v)
+  | .error _ => .error (.decode "BeaconState")
+
+/-- `runUpgrade` (the `fork` format, Gloas→Heze): decode the Gloas pre-state at preset `P`,
+apply `upgradeToHeze` with the config's `HEZE_FORK_VERSION`, return the Heze post root. A
+pure field copy; no onboarding / PTC recompute (builders are already onboarded in the Gloas
+state). -/
+private def runUpgradeImpl (P : Preset) (forkVersion : Version) (preBytes : ByteArray) :
+    Except (RunError StateTransitionError) ByteArray :=
+  letI : Preset := P
+  letI : HasherTag := fastHasherTag
+  match SSZ.deserialize (T := @Gloas.BeaconState P) preBytes with
+  | .ok pre  => .ok (htr (upgradeToHeze forkVersion pre))
+  | .error _ => .error (.decode "Gloas BeaconState")
+
+/-- Decode a Heze `BeaconState` into a `FastBox`, or the runner's `decode` error. -/
+private def decodeState (P : Preset) (bytes : ByteArray) :
+    Except (RunError StateTransitionError) (SSZ.Box Sha256 (@Heze.BeaconState P)) :=
+  letI : Preset := P
+  match SSZ.FastBox.deserialize (T := @Heze.BeaconState P) bytes with
+  | .ok box  => .ok box
+  | .error _ => .error (.decode "BeaconState")
+
+/-- `runEpochSubstep`: dispatch the `epoch_processing/<handler>` name to its Heze substep,
+run it over the decoded Heze pre-state, return the post root. Every substep is the Gloas
+one re-instantiated over Heze. -/
+private def runEpochSubstepImpl (P : Preset) (C : Config) (step : EpochStep) (preBytes : ByteArray) :
+    Except (RunError StateTransitionError) ByteArray := do
+  let box0 ← decodeState P preBytes
+  letI : Preset := P
+  letI : Config := C
+  letI : HasherTag := fastHasherTag
+  letI : CryptoBackend := CryptoBackend.realBackend
+  let action : EStateM StateTransitionError (SSZ.Box Sha256 (@Heze.BeaconState P)) Unit :=
+    match step with
+    | .slashingsReset               => processSlashingsReset
+    | .randaoMixesReset             => processRandaoMixesReset
+    | .eth1DataReset                => processEth1DataReset
+    | .historicalSummariesUpdate    => processHistoricalSummariesUpdate
+    | .participationFlagUpdates     => processParticipationFlagUpdates
+    | .justificationAndFinalization => processJustificationAndFinalization
+    | .inactivityUpdates            => processInactivityUpdates
+    | .rewardsAndPenalties          => processRewardsAndPenalties
+    | .registryUpdates              => processRegistryUpdates
+    | .slashings                    => processSlashings
+    | .effectiveBalanceUpdates      => processEffectiveBalanceUpdates
+    | .pendingDeposits              => processPendingDeposits
+    | .pendingDepositsChurn         => processPendingDeposits
+    | .pendingConsolidations        => processPendingConsolidations
+    | .builderPendingPayments       => processBuilderPendingPayments
+    | .syncCommitteeUpdates         => processSyncCommitteeUpdates
+    | .proposerLookahead            => processProposerLookahead
+    | .ptcWindow                    => processPtcWindow
+  RunError.ofSpec (runToRoot box0 action)
+
+/-- `runRewards`: the four reward-delta blobs computed by the inherited Heze delta
+functions. The `Deltas` container is Fulu's (fork-agnostic). -/
+private def runRewardsImpl (P : Preset) (C : Config) (preBytes : ByteArray) :
+    Except (RunError StateTransitionError) (Array ByteArray) := do
+  let state ← decodeState P preBytes
+  letI : Preset := P
+  letI : Config := C
+  letI : HasherTag := fastHasherTag
+  let mkDeltas : Array Gwei × Array Gwei → ByteArray := fun rp =>
+    SSZ.serialize ({ rewards := sszOfArray rp.1, penalties := sszOfArray rp.2 } : Fulu.Deltas)
+  let n := (sszGet state validators).size
+  let zeros := Array.replicate n (0 : Gwei)
+  RunError.ofSpec do
+    let d0 ← liftErr (getFlagIndexDeltas state 0)
+    let d1 ← liftErr (getFlagIndexDeltas state 1)
+    let d2 ← liftErr (getFlagIndexDeltas state 2)
+    pure #[mkDeltas d0, mkDeltas d1, mkDeltas d2, mkDeltas (zeros, getInactivityPenaltyDeltas state)]
+
+/-- Decode a plain (non-boxed) SSZ operation value. -/
+private def decodeOp (T : Type) [SizzLean.SSZRepr T] (b : ByteArray) :
+    Except (RunError StateTransitionError) T :=
+  match SSZ.deserialize (T := T) b with
+  | .ok v    => .ok v
+  | .error _ => .error (.decode "operation")
+
+/-- `runOperation`: dispatch every operation handler over the decoded Heze pre-state. Each
+handler is the Gloas one re-instantiated over Heze. -/
+private def runOperationImpl (P : Preset) (C : Config) (kind : OpKind)
+    (preBytes opBytes : ByteArray) (cmeta : CaseMeta) : Except (RunError StateTransitionError) ByteArray := do
+  let box0 ← decodeState P preBytes
+  letI : Preset := P
+  letI : Config := C
+  letI : HasherTag := fastHasherTag
+  letI : CryptoBackend := CryptoBackend.forBlsSetting cmeta.blsSetting
+  let dispatch : Except (RunError StateTransitionError) (EStateM StateTransitionError (SSZ.Box Sha256 (@Heze.BeaconState P)) Unit) :=
+    match kind with
+    | .proposerSlashing       => (decodeOp (@ProposerSlashing P) opBytes).map processProposerSlashing
+    | .attesterSlashing       => (decodeOp (@AttesterSlashing P) opBytes).map processAttesterSlashing
+    | .attestation            => (decodeOp (@Attestation P) opBytes).map processAttestation
+    | .payloadAttestation     => (decodeOp (@PayloadAttestation P) opBytes).map processPayloadAttestation
+    | .executionPayloadBid    => (decodeOp (@SignedExecutionPayloadBid P) opBytes).map processExecutionPayloadBid
+    | .parentExecutionPayload => (decodeOp (@BeaconBlock P) opBytes).map processParentExecutionPayload
+    | .blockHeader            => (decodeOp (@BeaconBlock P) opBytes).map processBlockHeader
+    | .withdrawals            => .ok processWithdrawals
+    | .voluntaryExit          => (decodeOp (@SignedVoluntaryExit P) opBytes).map processVoluntaryExit
+    | .voluntaryExitChurn     => (decodeOp (@SignedVoluntaryExit P) opBytes).map processVoluntaryExit
+    | .blsToExecutionChange   => (decodeOp (@SignedBLSToExecutionChange P) opBytes).map processBlsToExecutionChange
+    | .depositRequest         => (decodeOp (@DepositRequest P) opBytes).map processDepositRequest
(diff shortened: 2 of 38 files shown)


packages/EthCLSpecs/EthCLSpecs/Heze/Block.lean
```diff
@@ -0,0 +1,23 @@
...
+set_option autoImplicit false
+
+open EthCLLib.Spec
+open EthCLSpecs.Fulu
+
+namespace EthCLSpecs.Heze
+
+inherit BeaconBlockBody
+inherit BeaconBlock
+signedwrapper SignedBeaconBlock wraps BeaconBlock
```
leolara: Doesn't it work if we inherit SignedBeaconBlock ? if that if so, we should consider it a bug and fix it in another PR

leolara:
@IvanAnishchuk Heze includes changes to the fork choice that I don't see here, does it pass all fork choice tests in Heze?

IvanAnishchuk:
Pushed a big update, so the diff you reviewed is now well behind the branch. It's moved from the
minimal-cut snapshot the PR opened with to a conformance-complete and spec-complete Heze fork.
What changed:

The spine is wired now. EpochProcessing / Operations / Withdrawals / Transition / ForkChoice inherit
their Gloas counterparts over Heze's byte-identical twin types, and the fork-interface drives every
runner against them (the transition spine used to be stubbed as xfail). Every in-scope heze vector
passes: minimal 6770 / 0 failed, mainnet 997 / 0, zero xfail, all 137 `fork_choice` cases
(84 minimal + 53 mainnet) green.

`SignedBeaconBlock` inherits cleanly, so the old `signedwrapper` (the minimal-cut placeholder) is
gone. The FOCIL beacon-chain helpers (`get_inclusion_list_committee`,
`is_valid_inclusion_list_signature`) are in.

The FOCIL fork-choice layer is spec-complete: `InclusionListStore`, `on_inclusion_list`,
`get_inclusion_list_transactions`, the satisfaction gate, `get_inclusion_list_due_ms`, and the
`should_extend_payload` / `on_execution_payload_envelope` overrides. Alpha.11 ships `ssz_static`
vectors for the two new containers (they pass) but nothing that drives FOCIL behavior, so this
layer is spec-mirrored branch-for-branch and pinned with build-enforced `#guard`s /
`native_decide`; its deliberate divergences are written up in
`packages/EthCLSpecs/docs/DISCREPANCIES.md`. The EL verdict (`is_inclusion_list_satisfied`) sits
behind a small framework seam (`[ExecutionEngine]`, optimistic default), and a refuting test
instance drives the branch no vector reaches.

Two fixes reach outside the Heze layer. The fork-choice step interpreters were silently passing an
unmodeled `get_proposer_head` check through a catch-all; all three now match exhaustively
(unmodeled steps report `todo`, an xfail, and a new step type is a build error), and the harness
gives an unrecognized step or check key, nested ones included, the same treatment. And the
`pyspec_server` CLI errors on unknown fork or preset names instead of silently serving Fulu.

The description has the full picture, including the known limitation (the per-fork interpreter
clone) and a collapsed follow-ups list with checkboxes if you want to triage any of it.

packages/EthCLSpecs/EthCLSpecs/Gloas/Operations.lean
```diff
@@ -418,36 +389,41 @@ forkdef settleBuilderPayment (paymentIndex : Nat) : StateTransition Unit := do
...
-  -- A self-build sentinel carries no value or signature; a real builder must be
-  -- active, funded, and the signer of the bid.
+  -- A self-build sentinel carries no value or signature; a real builder must be active,
+  -- a payload builder (`PAYLOAD_BUILDER_VERSION`), funded, and the signer of the bid.
   if builderIndex == Const.builderIndexSelfBuild then
     assert (amount == 0)
     assert (signedBid.signature == Const.g2PointAtInfinity)
   else
     assert (isActiveBuilder state builderIndex)
+    assert ((sszGet state builders[builderIndex.toNat]!).version == Const.payloadBuilderVersion)
```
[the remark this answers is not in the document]
IvanAnishchuk: Good catch on an unchecked decoded index. One correction to the mechanism first, then what I'd do about it.

The pinned v1.7.0-alpha.11 spec has no bounds check here either: `process_execution_payload_bid` goes straight to `assert is_active_builder(state, builder_index)`, and `is_active_builder` reads `state.builders[builder_index]` with no guard, relying on Python's `IndexError` to invalidate the block. The Lean mirrors that structure. And `xs[i]!` in Lean doesn't raise: out of bounds it prints a panic line and returns the `Inhabited` default. So for this handler an out-of-range `builderIndex` yields a default builder, and the next `assert (isActiveBuilder …)` rejects it (default `withdrawableEpoch = 0 ≠ FAR_FUTURE_EPOCH`). The block still rejects cleanly, through default semantics plus stderr noise, with no exception and no false pass.

The underlying point holds though. This is implicit safety, and the codebase already has the explicit idiom (`assertH (idx.toNat < (sszGet state validators).size)` for the proposer index in proposer slashing; the new Heze inclusion-list-signature check in this branch guards its validator index the same way, returning false when out of range). The `!`-on-decoded-index pattern is systemic in the Gloas handlers (`isActiveBuilder`, `canBuilderCoverBid`, `verifyExecutionPayloadBidSignature`, the withdrawal reads), so the right fix is one consistent hardening pass across those sites plus a regression test that an out-of-range index rejects cleanly with no panic line, rather than a one-off assert that diverges from the spec's structure here. No alpha.11 vector feeds an out-of-range builder index, so it's latent. I'm tracking it as a dedicated follow-up rather than folding it into this fork-addition PR.

packages/EthCLSpecs/EthCLSpecs/Heze/Block.lean
IvanAnishchuk: Yes, it inherits cleanly. The branch replaces the old `signedwrapper` with `inherit SignedBeaconBlock` (`Heze/Block.lean:22`); EIP-7805 touches nothing in the block, so `BeaconBlockBody` / `BeaconBlock` / `SignedBeaconBlock` all inherit verbatim and the vectors are identical. No framework bug behind it, the `signedwrapper` was just the minimal-cut placeholder I opened with, so I folded the swap in here rather than filing a separate PR.

IvanAnishchuk:
Yes. Every heze `fork_choice` case passes, 84 minimal + 53 mainnet (the full heze sweep is 6770 passed / 0 failed / 0 xfail at minimal, 997 / 0 / 0 at mainnet). At alpha.11 EIP-7805's fork-choice additions (`on_inclusion_list`, the `InclusionListStore`, the satisfaction gate) ship no behavioral vectors: the corpus carries zero `on_inclusion_list` cases under heze (the two new containers do get full `ssz_static` suites, which pass). The two FOCIL overrides the vectors do reach, `should_extend_payload` and `on_execution_payload_envelope`, run on all of them and stay green.

On the changes you didn't see: the branch has moved since that review, and the fork-choice work is now in the diff, the inherited spine that drives every runner plus the FOCIL fork-choice layer itself. Since no vector reaches that layer yet, it's mirrored branch-for-branch against the pin and held by build-enforced `#guard` / `native_decide` pins, with its deliberate spec-text divergences recorded in `packages/EthCLSpecs/docs/DISCREPANCIES.md`. When FOCIL vectors ship, the handlers and the step decoding are already wired to receive them: an unknown step or check key reports `unsupported` instead of vanishing.

leolara:
so can you put it ready for review?

packages/EthCLSpecs/EthCLSpecs/Heze/Interface.lean
```diff
@@ -0,0 +1,402 @@
...
+      store := (← checkStepValidity store valid outcome)
+    | .payloadAttestationMessage bytes valid =>
+      let outcome := decodeStepOr (α := @Heze.PayloadAttestationMessage P) bytes "ptc message" fun msg =>
+        runOn store (onPayloadAttestationMessage (map := hashMap) msg false : EStateM StoreTransitionError (Store hashMap) Unit)
+      store := (← checkStepValidity store valid outcome)
+    | .checkHead root slot =>
+      let head := getHead store
+      assert (head.root == root)
+      let headSlot := match FcMap.lookup store.blocks head.root with | some b => b.slot.toNat | none => 0
+      assert (headSlot == slot)
```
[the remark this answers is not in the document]
IvanAnishchuk: Known one, it's on the follow-ups checklist in the description (the checkHead missing-block fallback item). The arm is a verbatim restatement of the same idiom in the Fulu and Gloas interpreters, the per-fork clone the description flags as the known limitation, so I'd rather harden all three together than diverge Heze here. It's also not reachable through a valid store: getHead walks from justifiedCheckpoint.root, which every handler keeps in blocks. And the pyspec-side check reads store.blocks[head].slot, so the same corrupt store would KeyError loudly there, no vector can encode the vacuous pass. A cross-fork missingKey throw is queued as a follow-up; happy to fold it into this PR instead if you'd rather have it now.

leolara:
## File organization: place the FOCIL code by concern, drop the feature files

`SPECS_ARCHITECTURE.md` §3.2 organizes state operations by concern, with no catch-all and no per-feature file. ePBS is the standing precedent. Its state-transition functions live in `Operations.lean`, `Withdrawals.lean`, and `EpochProcessing.lean`; its fork-choice functions live in `ForkChoice.lean`; only its containers get type-named files under `Containers/` (`Builder`, `PayloadBid`, `PayloadAttestation`, …). There is no `EPbs.lean`. A cross-cutting feature dissolves into the concerns it touches.

`Focil.lean` and `InclusionList.lean` are feature files, so they read against that grain. FOCIL is a cross-cutting feature exactly like ePBS, so its functions should land in concern files. Heze creates a concern file only for the genuinely-new functions and inherits everything reused, the same diff shape every other Heze file already follows.

### Target placement

| Declaration | Goes to | Layer |
|---|---|---|
| `getInclusionListCommittee` (+ its `cyclicSample` helper and the two `#guard` pins) | **`Heze/Committees.lean`** (new) | state (`state_section`) |
| `isValidInclusionListSignature` | **`Heze/Signing.lean`** (new) | state (`state_section`) |
| `InclusionListStore` forkstruct, `InclusionListStore.empty`, `collectInclusionListTransactions`, `processInclusionList`, `getInclusionListTransactions`, and their pins | **`Heze/ForkChoice.lean`** (already the FOCIL fork-choice home) | fork choice (`fork_choice_section`) |
| `isPayloadInclusionListSatisfied`, `shouldExtendPayload`, `isInclusionListSatisfied`, `recordPayloadInclusionListSatisfaction`, `onExecutionPayloadEnvelope`, `getForkchoiceStore`, `getInclusionListDueMs`, `onInclusionList`, the re-declared `Store` | already in `Heze/ForkChoice.lean` | fork choice |

After the move, `Heze/Focil.lean` and `Heze/InclusionList.lean` are deleted.

### Why each row

- **`getInclusionListCommittee` is a beacon-committee accessor.** Its concern is committees. Heze inherits `Committees` from Fulu and never re-declares it, so `Heze/Committees.lean` is a new concern file holding just this one new accessor; every other committee helper stays inherited. `cyclicSample` and its `#guard`s are the accessor's private machinery, so they travel with it.

- **`isValidInclusionListSignature` is a signature predicate.** Its concern is signing, so it goes in `Heze/Signing.lean`, a new concern file for the one new predicate, with the rest of the signing surface inherited. `blsVerifySigned` stays the `[CryptoBackend]` trust boundary as before.

- **`InclusionListStore` is fork-choice state, not an SSZ container.** It belongs with fork choice, next to the `Store` forkstruct that embeds it and the handlers that drive it (`onInclusionList`, `onExecutionPayloadEnvelope`, `recordPayloadInclusionListSatisfaction`, and the rest already in `ForkChoice.lean`). This matches how the fork-choice `Store` itself lives in `ForkChoice.lean` rather than under `Containers/`. Ordering constraint: declare `InclusionListStore` above the re-declared `Store` that carries it as a field. The store-logic pins move with the helpers they pin. Merged, `Heze/ForkChoice.lean` lands around 40 KB, in line with `Gloas/ForkChoice.lean` at 43 KB.

The result matches the ePBS shape: containers named per type under `Containers/`, every state and fork-choice function in the concern file for its primary effect, and a new file created only where Heze genuinely adds a function.

IvanAnishchuk:
Done, placed by concern as you laid out.

- `getInclusionListCommittee` (with `cyclicSample` and its pins) → `Heze/Committees.lean`
- `isValidInclusionListSignature` → `Heze/Signing.lean`
- `InclusionListStore` with its helpers and pins → `Heze/ForkChoice.lean`, declared above the `Store` that carries it

`Focil.lean` and `InclusionList.lean` are gone. Matches the Fulu `Committees`/`Signing` concern files and the ePBS shape per `SPECS_ARCHITECTURE.md` §3.2.

Build green, Heze conformance sweep unchanged at 6770 passed (minimal).

leolara:
please, rebase on the current main, and perhaps squash into one commit where there is no Claude attribution

IvanAnishchuk:
Done :saluting_face:

leolara (approved):

leolara:
@IvanAnishchuk I approved it but it would be better if the documentation is more clear.

In DISCREPANCIES.md, if those are implementation differences and not differences in the tests results from pyspec perhaps it should be moved to IMPLEMENTATION_NOTES.md.

But, also the text should be more clear, use more common jargon and clearer sentences, it feels very compressed.

Also, the comment before certain functions the text is difficult to follow.

IvanAnishchuk:
Thanks! Addressed in f8a6731.

You were right about DISCREPANCIES.md: those entries were implementation differences,
not test-result discrepancies. They now live in IMPLEMENTATION_NOTES.md as a per-fork
"Heze diff" section (one entry per function: spec behavior, Lean behavior, why no
vector can tell, where it's pinned), and DISCREPANCIES.md is back to being the narrow
spec-vs-vector ledger it says it is.

On the prose: the dense paragraphs became per-function entries, the invented
shorthand is gone, and house terms now carry a short gloss or a pointer at first use.
The hardest function comments (the inclusion-list store, the satisfaction gate, the
pin sections) were rewritten to state the point first, with each design rationale
told once and referenced from everywhere else instead of retold.

The pass also shook out some drift while at it: a couple of wrong references, and
stale two-fork / alpha.10 text in the README and Justfile. If anything still reads
dense, point at it and I'll take another pass.

packages/EthCLSpecs/docs/IMPLEMENTATION_NOTES.md
```diff
@@ -566,6 +596,96 @@ payload-availability bit), and `process_epoch` (builder-pending-payments and
 `process_ptc_window` last). The `transition` format folds pre-fork Fulu blocks, applies
 `upgradeToGloas` plus onboarding at the boundary, then folds post-fork Gloas blocks.
 
+## Heze diff
+
+`EthCLSpecs.Heze` is `fork Heze from Gloas` plus EIP-7805 (FOCIL). At alpha.11
```
leolara, line 601: This is not relevant, Heze changes are small now, so that is ok, not something we need to think so much about.

packages/EthCLSpecs/docs/IMPLEMENTATION_NOTES.md
```diff
@@ -566,6 +596,96 @@ payload-availability bit), and `process_epoch` (builder-pending-payments and
...
+the fork is a thin diff. EIP-7805 changes no state-transition substep, so the whole
+spine is the Gloas spine re-elaborated over Heze types, and the additions are two
+containers (`InclusionList`, `SignedInclusionList`), one committee accessor
+(`get_inclusion_list_committee`), one signature predicate
+(`is_valid_inclusion_list_signature`), and the fork-choice inclusion-list layer (the
+`InclusionListStore`, the satisfaction gate, and the `on_inclusion_list` handler).
+
+**Vector coverage.** FOCIL ships no behavioral conformance vector at the alpha.11 pin.
+The two new containers pass their full `ssz_static` suites, and the
```
leolara, line 610: This is relevant, vector coverage is not sufficient.

packages/EthCLSpecs/docs/IMPLEMENTATION_NOTES.md
```diff
@@ -566,6 +596,96 @@ payload-availability bit), and `process_epoch` (builder-pending-payments and
...
+- **`is_inclusion_list_satisfied`** (`heze/fork-choice.md:54-62`) defers its verdict to
+  `ExecutionEngine.is_inclusion_list_satisfied`, an Engine-API call against an external
+  execution client. This harness has no execution layer, so the call goes through the
+  `[ExecutionEngine]` typeclass (the Engine seam above), whose default instance answers
+  the constant `true`. That default is the residual execution-layer trust boundary of
+  the FOCIL gate. No conformance vector reaches the discriminating `false` branch; the
+  `pinRecordRefuted` `native_decide` example drives it end-to-end under a locally
+  substituted refuting engine instance.
+- **Two map reads default to `false` where the spec assumes the key present.**
+  `is_payload_inclusion_list_satisfied` (`heze/fork-choice.md:199-212`) opens with
```
leolara, line 634: We have `assert` why we cannot add it?

packages/EthCLSpecs/docs/IMPLEMENTATION_NOTES.md
```diff
@@ -566,6 +596,96 @@ payload-availability bit), and `process_epoch` (builder-pending-payments and
...
+  `assert root in store.payload_inclusion_list_satisfaction`; a pure `Bool` function
+  cannot throw, so the Lean reads a missing key as `false` (`lookupD`). A missing key
+  is unreachable when the spec's own invariant holds: `on_execution_payload_envelope`
+  writes `payloads` and `payload_inclusion_list_satisfaction` together, so the key is
+  present whenever `root ∈ payloads`. The `timeliness` read in
+  `get_inclusion_list_transactions` (`heze/inclusion-list.md:105-114`) is the same
+  pattern, with `process_inclusion_list` writing the stored list and its timeliness
+  entry together.
+- **`record_payload_inclusion_list_satisfaction`** (`heze/fork-choice.md:180-193`)
+  reads `Slot(state.slot - 1)`. On a slot-0 state the pyspec raises the uint64
```
leolara, line 644: We need to have the same behavoiur as the pyspecs, if we think they are wrong we can talk to them.

packages/EthCLSpecs/docs/IMPLEMENTATION_NOTES.md
```diff
@@ -566,6 +596,96 @@ payload-availability bit), and `process_epoch` (builder-pending-payments and
...
+  underflow, which invalidates the whole `on_execution_payload_envelope` call with the
+  store unmodified; the Lean guards the underflow, records the verdict over an empty
+  required set, and proceeds. Only an envelope whose target block state sits at slot 0
+  (the genesis anchor) could tell the difference, and no vector produces one.
+- **`should_extend_payload`** (`heze/fork-choice.md:221-236`) opens with
+  `assert store.blocks[root].slot + 1 == get_current_slot(store)`; the Lean is a total
```
leolara, line 650: again, we can assert, and we can convert it to a monad action to make it possible, if it throws in the pyspec it has to be a monad action that can throw

packages/EthCLSpecs/docs/IMPLEMENTATION_NOTES.md
```diff
@@ -566,6 +596,96 @@ payload-availability bit), and `process_epoch` (builder-pending-payments and
...
+
+Two model-structure choices carry no behavioral difference on any reachable input:
+
+- **`cyclicSample`** (`heze/beacon-chain.md:95-110`) resamples the committee as
+  `indices[i % len(indices)]`, which raises `ZeroDivisionError` on an empty
+  concatenation. The Lean read is total (`i % 0 = i`, then the `getD` default), and a
+  real beacon chain never reaches a zero-active-validator slot, so the branch is
+  unreachable (`EthCLSpecs/Heze/Committees.lean`).
+- **The `InclusionListStore` rides inside the fork-choice `Store` as a field.** The
+  spec keeps it as a process-lifetime singleton reached through
```
leolara, line 680: This is ok

leolara:
@IvanAnishchuk all these things that in pyspec can throw but here are pure functions should be changed to monad actions, which can throw, and can have `assert` also you can use the assert as proof for an indexing access.

For example: https://github.com/etheorem/etheorem/blob/main/packages/EthCLSpecs/EthCLSpecs/Fulu/Operations.lean#L69-L70

IvanAnishchuk:
This is the right call, and it's how we want the code to read. Everything pyspec can throw becomes a throwing monad action here, `assert` for the asserts, with the assert doubling as the bound for the indexing that follows.

The catch is blast radius. Applying it faithfully everywhere reaches well past this PR's surface, so folding the whole sweep into the Heze PR would roughly double its diff and pull in scope it should not carry. I put it in a stacked follow-up instead (#22), done as a full Gloas + Heze + Fulu sweep rather than only the flagged Heze sites. The commits stay granular on purpose, one mechanism each, so any part that is not right can be reverted or reworked on its own.

What the follow-up does:
- plain-`Dict` reads `getOrThrow` (`missingKey`): the Gloas `blocks` / `blockStates` / `checkpointStates` / `unrealizedJustifications` reads, the Heze `timeliness` read, the `is_payload_inclusion_list_satisfied` and `should_extend_payload` gates;
- dropped `assert`s are back as `assert`: the `should_extend_payload` slot equation, the `get_block_root_at_slot` recency assert, the `get_proposer_head` boost assert;
- uncaught `uint64` faults throw a distinct arithmetic reject: pyspec raises `ValueError` / `ZeroDivisionError` there, which the runner does not catch (only `AssertionError` / `IndexError`), so they stay out of the caught set. Two sites, the `record_payload_inclusion_list_satisfaction` slot-0 underflow (`Slot(state.slot - 1)`) and the `get_inclusion_list_committee` empty-committee `indices[i % 0]`;
- wire-index reads throw instead of masking as a default: `is_valid_inclusion_list_signature` reads `state.validators[validator_index]` through the safe indexed read (`outOfBounds`).

`getHead` and the weight/reorg walk thread the throwing monad end to end (`fuelLoop` in place of the pure `fuelIterate` for the block-reading walks), across all three forks. DefaultDict reads (`equivocators`, `inclusion_lists`, the payload-timeliness votes) stay defaults, matching the spec's `defaultdict`. Conformance stays green at both presets on all three forks.

Happy to keep auditing the rest of the codebase the same way and convert the remaining vectorless branches that still mask where the spec raises. One example already on the list: `get_balance_after_withdrawals`' `withdrawn` sum wraps on `uint64` overflow, where pyspec's `sum()` raises.

On `assertH`-as-index-proof: used where a later step needs the bound as a hypothesis. For `is_valid_inclusion_list_signature` I used the framework safe read (`sszGetIdx` → `outOfBounds`) rather than `assertH … ; xs[i]'h.down`, since nothing downstream needs the proof term and the reject is the same. Happy to switch it to `assertH` if you'd rather keep that idiom uniform.

packages/EthCLSpecs/docs/IMPLEMENTATION_NOTES.md
IvanAnishchuk: Added in #22, both throw now. `is_payload_inclusion_list_satisfied` asserts `root in store.payload_inclusion_list_satisfaction` and throws on a miss (`StoreTransition Bool`); the `timeliness` read in `get_inclusion_list_transactions` is a plain `Dict`, so it `getOrThrow`s (`missingKey`) in place of the old `lookupD false`. IMPLEMENTATION_NOTES updated.

packages/EthCLSpecs/docs/IMPLEMENTATION_NOTES.md
IvanAnishchuk: Matched in #22. `record_payload_inclusion_list_satisfaction` throws on the `Slot(state.slot - 1)` slot-0 underflow instead of recording over an empty required set. pyspec raises an uncaught `ValueError` there, so it throws the arithmetic reject, which the runner does not treat as an expected rejection.

packages/EthCLSpecs/docs/IMPLEMENTATION_NOTES.md
IvanAnishchuk: Done in #22. `should_extend_payload` is a throwing `StoreTransition` now: it `getOrThrow`s `blocks[root]`, asserts the slot equation, and `getOrThrow`s `blocks[proposer_root]`. This is the general rule from your summary comment, applied here and across the Gloas base.

packages/EthCLSpecs/docs/IMPLEMENTATION_NOTES.md
IvanAnishchuk: You're right that it's thin. FOCIL ships no behavioral conformance vector at alpha.11. The containers pass their `ssz_static` suites, and the `on_execution_payload_envelope` vectors only drive the Heze overrides on the empty-inclusion-list, always-satisfied path shared with Gloas. The FOCIL-specific branches (committee, timeliness, satisfaction gate, `on_inclusion_list`, the record path) have the pinned spec text as oracle, with build-enforced pins in `Heze/ForkChoice` / `Heze/Committees`. Now that the reads throw (#22), the pins cover both the accept and the reject side (`pinRecordSatisfied` / `pinRecordRefuted` / `pinCommitteeThrows` / the `pinCollect` set). Which FOCIL branches do you want deeper coverage on? I can add pins wherever you point.

leolara:
Closing in favour of #22, which is stacked directly on this branch and carries both commits here unchanged. The fork-choice throw-faithfulness work continues there, including the `assert`/throw points raised in the review threads above.

Nothing from this PR is dropped. Review and merge happen on #22.

PR #13, opened 2026-07-06
fix(hazmat-kzg): build LeanHazmatKzg on macOS

Mouzayan:
## Summary
Adds macOS support for LeanHazmatKzg by:

- using the Mach-O `__DATA,__const` section for the embedded trusted setup,
- using the platform-appropriate linker syntax for `LeanHazmatBls`,
- exporting the embedded trusted setup symbols with Mach-O-compatible names,
- documenting the macOS-specific requirements in the surrounding comments.

Linux behavior is unchanged.

## Test plan
- [x] `lake build LeanHazmatKzg` — builds clean on macOS (Apple Silicon)
- [x] `lake build LeanHazmatKzgTests` — KAT round-trips pass via `native_decide`

Note: CI only runs on `ubuntu-latest`, so it exercises the unchanged Linux
branch; the macOS path above was verified locally since CI can't cover it.


Mouzayan:
Follow-up to the LeanHazmatBls/Kzg fix above. `EthCLLib` and `EthCLSpecs`
independently re-supply a libcrypto link flag (Lake doesn't propagate
`moreLinkArgs` across `require`), and that flag was Linux-only
(`-l:libcrypto.so.3`). This blocked `lake build EthCLSpecs` on macOS
entirely, which in turn blocks writing and checking proofs against
`EthCLSpecs` locally (I'm setting up an `EthCLSpecs/Proofs/` module and
need a working macOS build to iterate on it).

Switched both packages to the portable `-lcrypto` plus the Homebrew
`opt` search paths, same "list every platform's libdir, let the linker
skip what's missing" pattern already used for the Linux paths. Also
added a dated addendum to `hazmat-docs/PLAN.md`'s Stage 0 outcome note,
which had recorded the old flag as settled, and a small CONTRIBUTING.md
fix so the Quick Start's native-deps step matches what README.md
already documents for macOS.

Verified with a clean `lake build EthCLSpecs` on macOS (365/365 jobs)
and `just ethcl-test`, whose native_decide self-tests (BLS sign/verify,
KZG cell-proof-batch round-trips through the `[CryptoBackend]` FFI seam)
exercise the linked shared libraries at runtime, confirming this isn't
just a link-time fix.

packages/EthCLLib/lakefile.toml
```diff
@@ -6,9 +6,12 @@ defaultTargets = ["EthCLLib"]
...
 # as `extern_lib`s, only the system shared lib needs re-stating here.
-# Two `-L` paths cover the common Linux libdir layouts (Debian multiarch and
-# `/usr/lib64` on Gentoo / Fedora / RHEL); the linker ignores whichever is absent.
-moreLinkArgs = ["-L/usr/lib/x86_64-linux-gnu", "-L/usr/lib64", "-l:libcrypto.so.3"]
+# The `-L` paths cover the common libdir layouts across platforms (Debian
+# multiarch, `/usr/lib64` on Fedora/RHEL, and Homebrew's `opt` symlink on
+# both macOS architectures); the linker ignores whichever are absent.
+# `-lcrypto` (not `-l:libcrypto.so.3`) resolves via the unversioned symlink
+# every platform's `-dev`/Homebrew package provides, so it works on both.
+moreLinkArgs = ["-L/usr/lib/x86_64-linux-gnu", "-L/usr/lib64", "-L/opt/homebrew/opt/openssl@3/lib", "-L/usr/local/opt/openssl@3/lib", "-lcrypto"]
```
IvanAnishchuk, line 14: This is starting to grow and arguably becoming a little fragile. I wouldn't recommend changing anything now, still manageable but I'd say we should reconsider the practice if the list ever grows by two more libraries or platform-specific paths. I'm very curious which will be the first, something windows-specific or a new library?

IvanAnishchuk (approved):
This is probably the correct way of adding a few alternative lib locations (I started the pattern adding the second location) and we'll probably survive adding a couple more platforms added like this.

I just wanted to mention for future discovery that it's possible to make this dynamic using pkgconfig (that would discover library paths for linker based on system configuration, which is more correct and robust) but the cost is converting nice declarative very readable toml into much more dense lakefile.lean and using something like `let libs ← runPkgConfig #["--libs", "libcrypto"]`

Mouzayan:
Rebased onto the latest `upstream/main`.

leolara (approved):

PR #16, opened 2026-07-09
feat(consensus-proof-spike): prove the builder-index flag round-trip

Mouzayan:
## Summary

Adds a curated shortlist of Gloas / EthCLLib theorem candidates, then
develops the first proof: three `bv_decide` theorems establishing the
builder-index flag round-trip and tagging properties for
`convertBuilderIndexToValidatorIndex` / `toBuilderIndex` in
`EthCLSpecs.Gloas.Operations`.

## What changed

- `docs/CONSENSUS_PROOF_CANDIDATES.md` (new): curated theorem candidates
  covering Gloas's 46 overrides and 62 new functions, each accompanied by a
  short rationale. The `convertBuilderIndexToValidatorIndex` row is marked
  **In review**, pointing at this PR's proof.
- `EthCLSpecs/Proofs/BuilderIndex.lean` (new): three theorems, all closed with
  `bv_decide`, no mathlib.
  - `toBuilderIndex_convertBuilderIndexToValidatorIndex` and
    `convertBuilderIndexToValidatorIndex_toBuilderIndex`: the two conditional
    round-trip theorems, each requiring the input to satisfy the corresponding
    `isBuilderIndex` precondition.
  - `isBuilderIndex_convertBuilderIndexToValidatorIndex`: the unconditional
    companion fact, setting the flag always makes `isBuilderIndex` true.
- `EthCLSpecs/Proofs.lean` (new) + `EthCLSpecs.lean`: wires the new proof
  module into the library so it is built as part of `EthCLSpecs`.
- `docs/IMPLEMENTATION_NOTES.md`: new "Proofs" section that records that
  proofs currently begin in `EthCLSpecs.Proofs` (mathlib-free), with
  `EthCLProofs` remaining reserved for proofs that require mathlib, following
  the conditional design in `SPECS_ARCHITECTURE.md` §11.3.
- `docs/README.md`: links to the new candidates doc.

## Test plan

  - [x] `just ethcl-test`
- [ ] N/A: no spec-type or cache-behaviour changes, `just ethcl-pyspec` not run.
- [ ] N/A: no bench-measured hot path touched.
- [ ] N/A: no new SSZ shapes / cache cases.

## Risk / trust footprint

None. No axiom, `sorry`, `partial def`, `native_decide`, `unsafe` block, or
`@[extern]` declaration added. All three theorems close with `bv_decide`
(`Std.Tactic.BVDecide`), producing fully checked proof terms with no new
trusted code.

## Notes for reviewers

- The round trip is not unconditional. `convertBuilderIndexToValidatorIndex`
  sets the flag bit, `toBuilderIndex` unconditionally clears it, so composing
  them is only a no-op if the input already satisfies the corresponding
  `isBuilderIndex` precondition. Each theorem states its precondition via
  `isBuilderIndex`, the spec's own named predicate for the bit test, rather
  than the raw `&&&` expression it unfolds to.
- `IMPLEMENTATION_NOTES.md` documents that proofs currently live in
  `EthCLSpecs.Proofs`. The standalone `EthCLProofs` package remains reserved
  for proofs that require mathlib, matching the conditional design described
  in `SPECS_ARCHITECTURE.md` §11.3.

## LICENSE (LGPL version 3)


packages/EthCLSpecs/EthCLSpecs/Proofs/BuilderIndex.lean
```diff
@@ -0,0 +1,72 @@
...
+`bv_decide`, no mathlib needed.
+
+See `EthCLSpecs/docs/CONSENSUS_PROOF_CANDIDATES.md`, "New Gloas functionality".
+-/
+
+set_option autoImplicit false
+
+namespace EthCLSpecs.Proofs
+
+open EthCLSpecs.Fulu
```
leolara: I was confused about this, general open, I think here it is better to import explicitly, it gave the impression that we were doing more with Fulu that what this is actually doing.

packages/EthCLSpecs/docs/CONSENSUS_PROOF_CANDIDATES.md
```diff
@@ -0,0 +1,90 @@
...
+---
+
+## Gloas overrides
+
+Functions Gloas redeclares under a name that also exists in Fulu, so a Fulu-vs-Gloas diff
+is often the cleanest way to state the theorem.
+
+| Function                         | Location                            | Rationale                                                                                                                      |
+| -------------------------------- | ----------------------------------- | ------------------------------------------------------------------------------------------------------------------------------ |
+| `processSlot`                    | `Gloas/Transition.lean:33-53`       | Identical to Fulu's `processSlot` except one new bit-clear; the small diff makes this a natural preservation theorem           |
```
leolara: I am not clear about what this would prove? do you mean to prove that they are almost the same between forks. I am not sure we need to proof things between forks. Or perhaps there is something to gain here, I don't know.

I am not sure about how this document is organised, we don't need to find specific what differs between Gloas and Fulu, or only new things. We would like to prove all Gloas (even things that were the same in Fulu). We don't start from a Fulu that is already proven.

packages/EthCLSpecs/EthCLSpecs/Proofs/BuilderIndex.lean
Mouzayan: Makes sense. The broad open is misleading here. I’ll narrow it to the specific names the file actually uses.

packages/EthCLSpecs/docs/CONSENSUS_PROOF_CANDIDATES.md
Mouzayan: Good point, thanks. I’ll rework the document to focus on properties to prove about Gloas itself rather than comparisons with Fulu, and update the organization to reflect that.

Mouzayan:
@leolara Thanks again for the review! I addressed the comments and pushed the updates. Ready for another look whenever you have time.

leolara (approved):

PR #22, opened 2026-07-15
feat(ethclspecs): fork-choice throw-faithfulness sweep (get_forkchoice_store + FOCIL)

IvanAnishchuk:
The pure total functions that pyspec throws on (via `assert`, a plain-`Dict` `KeyError`, an `IndexError`, or a `uint64` over/underflow), which were modeled as silent totals, now throw faithfully through the effect monad across the Fulu, Gloas, and Heze fork choices. Answers the fork-choice faithfulness review on etheorem/etheorem#6, which is closed in favour of this branch: the Heze layer is stacked underneath here commit for commit, so merging this merges that.

## What changed

- Gloas fork-choice runs through the throwing `StoreTransition` monad. That covers the weight/head/reorg walk, the leaf plain-`Dict` reads, `should_extend_payload`'s slot assert, and `getHead` via `fuelLoop`.
- Heze FOCIL fork-choice reads throw where pyspec does. `is_payload_inclusion_list_satisfied`, the timeliness plain-`Dict` read, `record_payload_inclusion_list_satisfaction`'s slot underflow, the inclusion-list-signature `IndexError`, and the empty inclusion-list-committee `ZeroDivisionError`.
- `get_forkchoice_store` throws its opening anchor assert (`anchor_block.state_root == hash_tree_root(anchor_state)`) in all three forks, seeding through `Except StoreTransitionError`.
- A new reject class, `StateTransitionError.arithmetic`, for a `uint64` over/underflow. pyspec raises `ValueError` there, which the reference runner leaves uncaught (unlike the `AssertionError` / `IndexError` it does catch), so it is a distinct uncaught class rather than a caught `.assert`. The withdrawal validator/builder reads go through `sszGetIdx` (the faithful `IndexError`) in place of a guarding `assert`, and the `block_timeliness[root][INDEX]` list reads throw through a new `arrGetIdx`.
- The runner's `.block` step scores `on_block` alone against the step's `valid` flag, then replays the block's own attestations / attester-slashings only on the valid path, matching pyspec's `add_block`.

### Checked `uint64` arithmetic

`checkedAdd` / `checkedSub` / `checkedMul` reject on overflow where the bare operators wrap. Converted: the `get_forkchoice_store` time seed, `get_slots_since_genesis`, `compute_time_at_slot`, `time_into_slot`, `on_tick`'s per-slot catch-up, `is_finalization_ok`'s epoch gap, `get_proposer_head`'s slot comparisons, `process_attestation`'s inclusion delay, `process_execution_payload`'s timestamp product and sum, `get_ptc`'s window index, `get_balance_after_withdrawals`' accumulator, and the Gloas slot/epoch adds in `filter_block_tree`, `is_previous_slot_payload_decision` and `is_proposer_equivocation`.

`on_tick`'s catch-up runs on a new `fuelIterateM!`, which throws on fuel exhaustion rather than returning the accumulator.

All three operators raise through one helper, `throwArithmetic`, which also serves the spec's single non-operator arithmetic fault (`get_inclusion_list_committee`'s `indices[i % len(indices)]` on an empty committee, a `ZeroDivisionError`). No call site names the per-machine wrapper; the `[ErrorConv StateTransitionError E]` bound rides the fault in as `.arithmetic` on the state machine and `.transition (.arithmetic …)` on the store machine.

### Per-step-kind reject scoring

The reference's caught set differs by step kind: `expect_assertion_error` (`context.py:424-435`) catches `IndexError` for the attestation, attester-slashing, envelope and PTC-message steps, while `add_block` (`fork_choice.py:389`) catches `AssertionError` and `BlockNotFoundException` alone. `FcStepKind` carries that table and `checkStepValidity` reads it.

A store-side `missingKey` classifies as an uncaught fault rather than a likely bug: it is a bare-`Dict` `KeyError`, which the reference propagates, so it can never be an invalid vector's expected rejection. A decode failure scores as a failure rather than a rejection.

### Fixes found while reviewing

- **A decode failure no longer clean-passes an invalid step.** A fork-choice SSZ decode failure was thrown as `.assert`, an expected rejection, so a decoder or wrong-container bug on a `valid:false` step passed green. `StoreTransitionError.decodeFailure` stays out of `isExpectedRejection`.
- **A nested `todo` / `outOfScope` reports correctly**, routed through `RunError.classify`.
- **`getWeight` reads `store.blocks` inside the generator's short-circuit branches**, matching pyspec. Hoisting it made an empty-`latest_messages` store, exactly what `get_forkchoice_store` returns, raise `.missingKey` where the reference returns `Gwei(0)`.
- **A column sidecar the decoder cannot read rejects** rather than being dropped, which would let a `valid:false` step pass on a broken decoder.

### Review follow-ups

Five commits on top of the original stack, from Leo's review pass.

- **Fuel exhaustion rejects as `todo`, not `assert`** (`b69aa01`). `fuelIterateM!`'s bound belongs to this model rather than to spec text, so exhausting it is a work-queue item. As `assert` it sat in the caught set, which would have passed a `valid: false` step green on a bound we already know is wrong. Both call sites are `on_tick`, whose steps always run `expectedValid := true`, so no verdict moves. The class keeps it that way for the next caller.
- **One canonical raise for the arithmetic fault** (`fc68ce6`). Five sites built `.arithmetic` by hand, two of them naming the `.transition` wrapper themselves. The three `balanceAfterWithdrawals` copies and `record_payload_inclusion_list_satisfaction` were open-coding `checkedSub` and now call it; `throwArithmetic` covers the one genuine non-operator fault. Behavior-preserving, and `pinBalanceUnderflowThrows` still holds.
- **Three documentation corrections** (`258a7e9`, `c98cb4c`, `e24e79c`). The `cyclicSample` note still said the caller "asserts", naming the caught class where the guard is an uncaught `.arithmetic` throw. `betterOf` now records why its `foldlM` surfaces the same reject as the spec's `max`: an accumulator reached its position by surviving an earlier step, so its key already succeeded and cannot throw on the re-run. `getWeight` names its two deliberate non-conversions (the unchecked index is in range by construction, the `Gwei` sums defer with the weight path) where they previously lived only in this description.

## Faithfulness pins

Each throw whose reject branch no conformance vector reaches carries a build-enforced pin: `pinCommitteeThrows`, `pinRecordRefuted`, `pinAnchorRejects`, `pinBalanceUnderflowThrows`, and the collect/timeliness pins. The class-sensitive ones match on the reject constructor, so a regression from `.assert` to `.missingKey` breaks the build. `native_decide` where a hash is involved, kernel `#guard` otherwise.

`fuelIterateM!`'s fuel-out is pinned the same way in `Tests/FrameworkUtils.lean`, on both the constructor and `isExpectedRejection`, so a regression back to `.assert` breaks the build rather than turning a wrong loop bound into a green step.

## Conformance

All six sweeps pass-to-pass against baseline, zero failures:

| fork | minimal | mainnet |
|------|---------|---------|
| fulu | 5188 | 847 |
| gloas | 6573 | 1034 |
| heze | 6770 | 997 |

21409 total. Every converted throw is unreachable on a current conformance vector, so no verdict moves; the pins carry the regression guard.

The five review follow-up commits were verified on `just ethcl-test` (439 jobs green, the `native_decide` and `#guard` pins included) and `just ethcl-pyspec-smoke` (fulu 152, gloas 200, heze 201 passed, zero failures). The full sweep above has not been re-run since; only one of the five touches a live code path, and it is behavior-preserving.

Note for anyone reproducing the build: bare `lake build` at the repo root builds nothing and still exits 0, because the umbrella package declares no `lean_lib`. Use `just ethcl-test`.

<details>
<summary>Deferred findings (all filed as issues)</summary>

Real, but unreachable on any current conformance vector, or cleanup. Every one is now an issue, so none of it depends on this branch surviving the squash. The detail lives in the issues; this is the index.

| Issue | Site | What |
|---|---|---|
| etheorem/etheorem#43 | `Spec/Balances.lean`, the weight path | `getTotalBalance`'s fold wraps where the spec's `sum(...)` raises. The whole weight path rests on it and converts as one piece. |
| etheorem/etheorem#44 | state-transition `uint64` arithmetic | Bare `+` / `-` / `*` remains across epoch processing and the registry. Partly blocked on #41. |
| etheorem/etheorem#45 | `Fulu/Transition.lean` | `process_execution_payload`'s timestamp assert keeps the seconds form where the spec uses milliseconds. Value-equivalent at both presets. |
| etheorem/etheorem#46 | `Gloas/ForkChoice.lean` | `is_data_available` / `verify_and_notify_new_payload` are inline `true`. Migrate onto the `[ExecutionEngine]` seam. |
| etheorem/etheorem#47 | `PySpecTests/harness.py` | The `flagged` bug-smell marker is computed, sent as `bug!`, and discarded before any human sees it. |
| etheorem/etheorem#48 | `Gloas/Interface.lean` | `get_proposer_head` routes to `.todo` for Gloas. Fulu implements it. |
| etheorem/etheorem#49 | `Spec/Engine.lean` | FOCIL `is_inclusion_list_satisfied` is the optimistic default; only `pinRecordRefuted` reaches the `false` branch. |
| etheorem/etheorem#50 | `Heze/Signing.lean` | `isValidInclusionListSignature` carries no pin and no vector, so the EIP-7805 signature and domain path is unexercised. |
| etheorem/etheorem#41 | `Spec/Errors.lean` | The caught set is per step kind but still one policy per runner. Blocks `initiate_validator_exit`'s conversion to `checkedAdd`. |
| etheorem/etheorem#51 | `PySpecTests/harness.py` | `ServerClient` mutes Lean panics, cannot time out a wedged server, and does not reap the previous `Popen`. |
| etheorem/etheorem#52 | `EthCLSpecs` / `EthCLLib` | ~650 lines of `native_decide` pins ship in the precompiled libraries rather than the `*Tests` libraries the lakefiles designate. |
| etheorem/etheorem#53 | the three `Interface.lean` | The fork-choice step interpreter is triplicated across forks and kept in lockstep by hand. |
| etheorem/etheorem#54 | the three `Interface.lean` | The block-replay sub-action is labelled `.attestation` though it replays attester slashings too. Sub-issue of #53. |
| etheorem/etheorem#55 | `Gloas/ForkChoice.lean` | `betterOf` recomputes `getWeight` for both sides of each comparison. Performance only; `c98cb4c` records why it is otherwise unobservable. |
| etheorem/etheorem#56 | `SPECS_ARCHITECTURE.md` §7.2 | The fuel-versus-measure justification is recorded per section rather than per loop site. |

</details>

### DIFF
--- a/Justfile
+++ b/Justfile
@@ -10,6 +10,11 @@
 #
 # The pyspec recipes need a Python venv. Run `just setup-python` once first.
 
+# xdist worker count for the full pyspec sweeps; `auto` (one per core) is the
+# historical default. Override on memory-constrained machines:
+# `PYTEST_JOBS=2 just ethcl-pyspec-full`.
+pytest_jobs := env_var_or_default("PYTEST_JOBS", "auto")
+
 # List every recipe with its description
 default:
     @just --list --unsorted
@@ -251,24 +256,27 @@ ethcl-pyspec args="": _ensure-venv
 # xfail as the Phase-2 work-queue, so the run is green (exit 0) iff no in-scope
 # vector hits a bug-smell or a real mismatch. Mainnet / full sweep run on demand.
 
-# CI smoke gate: EthCLSpecs pyspec dev subset for both forks at minimal
+# CI smoke gate: EthCLSpecs pyspec dev subset for all three forks at minimal
 [group('ethcl')]
 ethcl-pyspec-smoke: _ensure-venv
     cd packages/EthCLSpecs/PySpecTests && {{ justfile_directory() }}/.venv/bin/python -m pytest -q --fork=fulu --subset=2
     cd packages/EthCLSpecs/PySpecTests && {{ justfile_directory() }}/.venv/bin/python -m pytest -q --fork=gloas --subset=2
+    cd packages/EthCLSpecs/PySpecTests && {{ justfile_directory() }}/.venv/bin/python -m pytest -q --fork=heze --subset=2
 
 # The complete in-scope sweep: every collected vector (`--subset=0`) for the
-# full matrix of {fulu, gloas} × {minimal, mainnet}, sharded across cores. The
-# two minimal forks finish quickly; the two mainnet forks are the long poles
+# full matrix of {fulu, gloas, heze} × {minimal, mainnet}, sharded across cores. The
+# three minimal forks finish quickly; the three mainnet forks are the long poles
 # (real-size SSZ + crypto). Each xdist worker holds its own warm `pyspec_server`.
 
-# Full EthCLSpecs pyspec sweep: {fulu,gloas} × {minimal,mainnet}, sharded across cores
+# Full EthCLSpecs pyspec sweep: {fulu,gloas,heze} × {minimal,mainnet}, sharded across cores
 [group('ethcl')]
 ethcl-pyspec-full: _ensure-venv
-    cd packages/EthCLSpecs/PySpecTests && {{ justfile_directory() }}/.venv/bin/python -m pytest -q --subset=0 -n auto --preset=minimal --fork=fulu
-    cd packages/EthCLSpecs/PySpecTests && {{ justfile_directory() }}/.venv/bin/python -m pytest -q --subset=0 -n auto --preset=minimal --fork=gloas
-    cd packages/EthCLSpecs/PySpecTests && {{ justfile_directory() }}/.venv/bin/python -m pytest -q --subset=0 -n auto --preset=mainnet --fork=fulu
-    cd packages/EthCLSpecs/PySpecTests && {{ justfile_directory() }}/.venv/bin/python -m pytest -q --subset=0 -n auto --preset=mainnet --fork=gloas
+    cd packages/EthCLSpecs/PySpecTests && {{ justfile_directory() }}/.venv/bin/python -m pytest -q --subset=0 -n {{ pytest_jobs }} --preset=minimal --fork=fulu
+    cd packages/EthCLSpecs/PySpecTests && {{ justfile_directory() }}/.venv/bin/python -m pytest -q --subset=0 -n {{ pytest_jobs }} --preset=minimal --fork=gloas
+    cd packages/EthCLSpecs/PySpecTests && {{ justfile_directory() }}/.venv/bin/python -m pytest -q --subset=0 -n {{ pytest_jobs }} --preset=minimal --fork=heze
+    cd packages/EthCLSpecs/PySpecTests && {{ justfile_directory() }}/.venv/bin/python -m pytest -q --subset=0 -n {{ pytest_jobs }} --preset=mainnet --fork=fulu
+    cd packages/EthCLSpecs/PySpecTests && {{ justfile_directory() }}/.venv/bin/python -m pytest -q --subset=0 -n {{ pytest_jobs }} --preset=mainnet --fork=gloas
+    cd packages/EthCLSpecs/PySpecTests && {{ justfile_directory() }}/.venv/bin/python -m pytest -q --subset=0 -n {{ pytest_jobs }} --preset=mainnet --fork=heze
 
 # ═════════════════════════════════════════════════════════════════════════
 # SizzLean — SSZ library
--- a/Justfile
+++ b/Justfile
@@ -229,7 +229,7 @@ _ensure-venv:
     fi
 
 # ═════════════════════════════════════════════════════════════════════════
-# EthCLSpecs — consensus-spec framework + Fulu / Gloas fork bodies
+# EthCLSpecs — consensus-spec framework + Fulu / Gloas / Heze fork bodies
 #
 # EthCLLib + EthCLSpecs (the consensus-spec framework + Fulu/Gloas bodies). The
 # `*Tests` libs carry the framework + spec `#guard` / `native_decide` self-tests
@@ -237,7 +237,7 @@ _ensure-venv:
 # building them fires the gates.
 # ═════════════════════════════════════════════════════════════════════════
 
-# EthCLLib + EthCLSpecs self-tests (framework + Fulu/Gloas spec gates)
+# EthCLLib + EthCLSpecs self-tests (framework + fork spec gates)
 [group('ethcl')]
 ethcl-test:
     lake build EthCLLib EthCLLibTests EthCLSpecs EthCLSpecsTests
@@ -252,7 +252,7 @@ ethcl-pyspec args="": _ensure-venv
     cd packages/EthCLSpecs/PySpecTests && {{ justfile_directory() }}/.venv/bin/python -m pytest -q {{ args }}
 
 # CI smoke gate for EthCLSpecs pyspec: the dev subset (a few cases per
-# handler) at minimal for both forks. Currently-green formats pass; the rest
+# handler) at minimal for all three forks. Currently-green formats pass; the rest
 # xfail as the Phase-2 work-queue, so the run is green (exit 0) iff no in-scope
 # vector hits a bug-smell or a real mismatch. Mainnet / full sweep run on demand.
 
@@ -288,7 +288,7 @@ ethcl-pyspec-full: _ensure-venv
 # boolean, the test-only containers); the EIP-7495 / 7916 / 8016 progressive /
 # stable / compatible forms are out of `SizzLean`'s universe and xfail. The
 # per-fork consensus-container `ssz_static` vectors run inside the EthCLSpecs
-# `ethcl-pyspec*` recipes (Fulu + Gloas), not here.
+# `ethcl-pyspec*` recipes (the fork bodies), not here.
 # ═════════════════════════════════════════════════════════════════════════
 
 # `SizzLeanTests.PendingListShrink` Cases 4/5/7 deliberately drive
--- a/README.md
+++ b/README.md
@@ -26,7 +26,7 @@ Upstream repository: <https://github.com/etheorem/etheorem>.
 ```
 LeanSha256 ─────────────┐
                         ├─→ SizzLean ──→ EthCLLib ──→ EthCLSpecs
-LeanHazmatSha256 ───────┘   (SSZ +       (consensus    (Fulu / Gloas
+LeanHazmatSha256 ───────┘   (SSZ +       (consensus    (Fulu…Heze
    (FFI SHA-256)            cache)        framework)    fork bodies)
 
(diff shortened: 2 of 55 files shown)


leolara:
Review note on `balanceAfterWithdrawals` (the Fulu, Gloas, and Heze copies): the `withdrawn` accumulator folds with a wrapping `UInt64` `+`, while pyspec's `sum(...)` (`capella/beacon-chain.md:368-378`) accumulates an unbounded int, so the underflow `ValueError` fires against the true sum. A true sum ≥ 2^64 wraps here, can pass the `withdrawn > bal` guard, and returns a wrong balance where pyspec raises. Unreachable on a well-formed state, and the PR description already defers it, but the three docstrings say "modeled faithfully" and `IMPLEMENTATION_NOTES.md` carries no entry. Recommend one of: accumulate in `Nat` and throw `.arithmetic` when the true sum exceeds the balance, or add the deferral to `IMPLEMENTATION_NOTES.md` so the surviving docs match the code.

leolara:
On the `IMPLEMENTATION_NOTES.md` rework: the file's intro defines the per-fork catalogue as "deliberate divergences from the spec text that no vector observes", but the Heze section now opens with "how it meets the spec" and mostly confirms faithful throws. Once a divergence is fixed, deletion is the record; a "throws to match" entry keeps a ghost of the old divergence, and each one duplicates an argument the code docstrings and pins already carry. Several also narrate before/after ("in place of the old `lookupD false`", "the former panicking `[i]!`"), history that belongs in the commits.

Suggested pruning:

- Heze diff section: drop the entries for the two map reads, `record_payload_inclusion_list_satisfaction`, `should_extend_payload`, `is_valid_inclusion_list_signature`, `get_forkchoice_store`, and the `get_inclusion_list_committee` paragraph. Keep the `[ExecutionEngine]` seam and `InclusionListStore`-in-`Store` entries (genuine divergences), and revert the section lead to divergences-only. The pin-regime intro above the list already says the rest once.
- Fork choice section: drop the "weight/head walk throws its plain-`Dict` reads faithfully" passage and the `verifyExecutionPayloadEnvelopeSignature` paragraph; same for the matching sentence in `DISCREPANCIES.md`. Keep the DAG-walk paragraph, the UInt64-wrap caveat, and the `checkStepValidity` caught-set paragraph. The caught set documents the reference harness rather than the spec, so it earns a doc home, though it needs one correction: the block-step wrapper catches only `AssertionError` / `BlockNotFoundException` (`fork_choice.py:389`), so "the reference runner's `AssertionError` / `IndexError` set" is too uniform across step kinds.

leolara:
Two Heze pins are class-blind and could be tightened. `pinAnchorRejects` checks `.toOption.isNone`, and the `pinPils … = none` case collapses every error branch to `none`, so both pin only *that* a throw happens, not *which* class. A regression from `.assert` (caught, expected rejection) to `.missingKey` (uncaught, bug-smell) at either site would still satisfy the pin and pass the build, even though that swap flips whether a `valid: false` vector scores as a faithful rejection or a failure, the exact distinction this PR is about.

Fix: match on the error constructor the way the Gloas / Fulu pins already do. For `pinAnchorRejects`, assert the result is specifically `.error (.assert _)` rather than `isNone`. For `pinPils`, map the `.assert` reject to a distinct marker instead of folding all errors to `none`, so the `pinPils true none` case pins `.assert` specifically. `pinCommitteeThrows` (which checks `!e.isExpectedRejection`) is already class-aware and can stay as the model. Test-only, no spec-code change.

leolara:
Stale row in the `Driver.lean` scoring table (`PySpecTests/Driver.lean:21`):

`| invalid | outOfBounds / missingKey reject (caught IndexError) | pass, flagged (bug-smell) |`

Two issues: this table documents the state-transition driver, whose reject type is `StateTransitionError`, which has no `missingKey` constructor (`missingKey` is a `StoreTransitionError`, the fork-choice store machine), so that case can't arise here. And the "caught IndexError" label lumps the two classes together, `outOfBounds` (`IndexError`) is caught, but a `missingKey` (`KeyError`) is never caught by the reference. Fix: drop `missingKey` from the row so it reads `outOfBounds reject (caught IndexError)`. Docstring-only.

IvanAnishchuk:
Fixed rather than deferred, and the premise needs a correction first: `sum(...)` does not accumulate an unbounded int. `Gwei` subclasses `int` and defines `__radd__`, so Python's reflected-operand rule makes the accumulator a `Gwei` from the first element on, and every add re-runs the bound check in `uint.__new__` (`remerkleable/basic.py:88-107`). Checked against the pinned remerkleable: `sum`, `+` and `*` on `uint64` all raise `ValueError` on overflow.

So pyspec raises during the accumulation, not at the subtraction, and the fold is the faithful place to fault. All three copies accumulate through `checkedAdd` now, which throws `.arithmetic` at that same operation. Nat accumulation would have landed the same reject class one step later, at the guard.

Dropped the `IMPLEMENTATION_NOTES` entry with it, since the divergence is gone.

The weight path still wraps and stays deferred, and its note is rewritten around the one thing it rests on: `getTotalBalance`'s fold. `committeeWeight` is that total over `SLOTS_PER_EPOCH`, `calculateCommitteeFraction` and `getProposerScore` are its products, and `getWeight`'s boost add takes an operand from them, so checking those while the total underneath wraps only moves a wrong number around. `getTotalBalance` and `getTotalActiveBalance` are pure `forkdef`s with 15 call sites across epoch processing, rewards, registry updates and the transition spine, so converting them propagates through every pure caller; that is the piece of work, not a handful of arithmetic sites.

`bpsDeadlineMs` came off the deferred list entirely: both factors are constants and the product peaks near 1.2e8, so it cannot reach the `uint64` bound at any preset.

IvanAnishchuk:
Pruned as listed, kept the `[ExecutionEngine]` seam and `InclusionListStore` entries, lead is divergences-only again. The caught-set paragraph carries your correction and `FcStepKind` holds the same table in code.

Two more doc claims failed against the tree while I was in there: `SPECS_ARCHITECTURE` 7.1 specified a pure fork-choice read layer (`getHead`, `getWeight` and `filterBlockTree` are all `StoreTransition`), and Fulu `ssz_static` was called out of scope in four places while it runs for real.

IvanAnishchuk:
Both tightened on the `pinCommitteeThrows` model: `pinPils` answers a `PilsVerdict` with `.assertReject` matching `.error (.assert _)`, and `pinAnchorRejects` matches the constructor instead of `isNone`. Fulu and Gloas pins got the same treatment.

IvanAnishchuk:
Also please let me know if you want me to squash this thing or something.

leolara (approved):

PR #23, opened 2026-07-15
feat(LeanImtPlus): implement authenticated SHA-256 tree and unified proofs

protocolwhisper:
## Summary

  This PR implements the executable SHA-256 LeanIMT+ tree lifecycle in pure Lean,
  following the commitment and proof format of
  [`vplasencia/leanimt-plus`](https://github.com/vplasencia/leanimt-plus).

  LeanIMT+ gives Ethereum and ZK protocols an authenticated data structure that
  can prove absence as well as presence. It combines LeanIMT’s dynamic depth and
  unpaired-node promotion with indexed leaves, producing compact
  non-membership proofs without the fixed depth and zero-hash padding commonly
  used by sparse Merkle trees.

  Etheorem previously exposed a helper that validated proof shape and recomputed
  a candidate root. This PR adds the tree operations, proof generation, and
  root-bound verification needed for nullifier tracking, credential revocation,
  allowlists, and differential validation between Lean and ZK circuits.

  ## What changed

  ### LeanImtPlus tree

  - Added insertion, batch insertion, tombstone removal, and in-place updates.
  - Added predecessor-link repair after removal.
  - Added dynamic Merkle depth and promotion of unpaired nodes.
  - Added active-leaf lookup, tree metadata, and SHA-256 root calculation.
  - Preserved physical insertion order and permanent tombstone slots.

  ### Proof generation and verification

  - Added unified membership and non-membership proof generation.
  - Added the committed root to `Proof`.
  - Changed `verify` to authenticate the recomputed path against that root.
  - Added `recomputeRoot` for circuit wrappers that bind the root separately.
  - Added `verifyProof` for canonical unified proofs.
  - Added typed errors for invalid ranges, malformed paths, invalid proof
    semantics, tombstone replay, and root mismatch.
  - Added canonical sibling-length, leaf-index, and circuit-depth checks.

  ### SHA-256 compatibility

  - Preserved the circuit’s 216-bit input range.
  - Preserved 27-byte big-endian encoding and low 216-bit digest truncation.
  - Preserved the tagged leaf commitment:

    ```text
    hash2(hash2(value, nextValue), 1)

  - Preserved two-child internal hashing and LSB-first path traversal.
  - Matched the canonical SHA-256 fixture from vplasencia/leanimt-plus.
  - Confirmed that inserting [10, 25, 7, 3, 41, 18] produces the reference
    root.

  ### Tests

  - Added membership proof coverage.
  - Added non-membership coverage below the minimum, between active leaves,
    above the maximum, and after removal.

  - Added insertion, removal, update, and tombstone lifecycle coverage.
  - Added rejection tests for root tampering, invalid ranges, malformed paths,
    invalid proof semantics, and tombstone replay.

  - Added mutation error coverage for zero values, duplicates, empty batches,
    and missing values.

  ### Repository integration

  - Added LeanImtPlus to the root build recipe.
  - Added LeanImtPlusTests to just test.
  - Added the just leanimt-plus-test recipe.
  - Included LeanIMT+ in the CI test job.
  - Added package documentation and updated the monorepo build documentation.

  ## Test plan

  - [ ] lake build (full monorepo, on the pinned toolchain) is green.
    This remains for CI or an environment containing all native
    dependencies.

  - [ ] Per-package test suites pass:
      - [ ] lake build LeanSha256Tests
      - [ ] lake build SizzLeanTests
      - [ ] just ethcl-test
      - [x] lake build LeanImtPlusTests
      - [x] cd packages/LeanImtPlus && lake build

    just ethcl-pyspec is green.
    Not applicable. Consensus types and SSZ cache behaviour remain
    unchanged.

  - [x] If this touches a bench-measured hot path: included
    before/after lake exe ssz_bench TSV (or relevant rows) in
    the PR description.
    Not applicable. LeanImtPlus is outside the existing SSZ benchmarks.

  - [x] New SSZ shapes / cache cases come with a native_decide or
    No SSZ shapes are added. LeanIMT+ roots and proof paths have
    native_decide tests.

  - [x] git diff --check is clean.
  - [x] The forbidden Lean token check is clean.

  ## Risk / trust footprint

  - New explicit axioms: none.
  - New sorry: none.
  - New partial def: none.
  - New unsafe blocks: none.
  - New @[extern] declarations: none.
  - New external dependencies: none.
  - Adds five native_decide invocations for concrete SHA-256 roots and paths.

  The native_decide tests follow the project convention for concrete hash
  computations and carry the standard compiler-evaluation trust commitment.

  The public proof API changes:

  - Proof now carries its committed root.
  - VerifyResult changes from Except VerifyError Nat to
    Except VerifyError Unit.

  - verify succeeds when the path reconstructs the committed root.
  - Callers that need the candidate root can use recomputeRoot.

  ## Notes for reviewers

  ### Why LeanIMT+ belongs in Etheorem

  Etheorem provides executable and formally analyzable Ethereum primitives in
  Lean. Authenticated non-membership is useful when accepting an operation
  depends on showing that an identifier has not appeared in a committed set.

  A privacy protocol can prove that a nullifier is absent before accepting a
  spend, then insert it to prevent reuse. A credential system can prove that an
  identifier is absent from a revocation registry. Allowlists and denylists
  similarly need compact proofs of presence or absence.

  Sparse Merkle trees support non-membership but commonly use fixed depth and
  zero-hash padding. LeanIMT+ uses indexed leaves for the sorted predecessor
  relation while retaining LeanIMT’s dynamic depth and promotion of unpaired
  nodes.

  This gives Etheorem a reusable authenticated-set primitive and exercises
  LeanSha256 as the foundation of a higher-level data structure.

  ### Security motivation

  The previous verify operation returned the root reconstructed from the
  supplied leaf and siblings without comparing it with a trusted commitment. A
  caller could interpret .ok computedRoot as successful authentication.

  The new Proof.root field makes this boundary explicit. A path that
  reconstructs a different root returns .error .rootMismatch.

  For non-membership, the verifier checks:

  leaf.value < value

  and:

  value < leaf.nextValue OR leaf.nextValue = 0

  The tombstone replay guard permits a zero-valued low leaf only at index zero,
  which is reserved for the sentinel.

  ### Reference and compatibility

  The implementation follows vplasencia/leanimt-plus at commit
  e95da64 (https://github.com/vplasencia/leanimt-plus).

  The following behavior matches the SHA-256 reference:

  - sentinel creation and physical insertion order;
  - sorted nextValue links and permanent tombstones;
  - in-place updates and odd-node promotion;
  - leaf and internal commitments;
  - compressed sibling paths;
  - membership and non-membership semantics;
  - committed roots.

  ### Current design trade-off

  The reference uses an AVL index for O(log n) predecessor lookup. The current
  Lean implementation scans physical slots, giving lookup O(n) complexity.

  The reference also caches Merkle levels and updates affected ancestors. The
  current implementation reconstructs levels when calculating a root or
  generating a proof.

  These choices preserve compatible commitments and proofs while focusing this
  PR on executable semantics and security. Follow-ups can add:

  - a persistent Lean AVL index;
  - cached Merkle levels and incremental ancestor updates;
  - differential property tests and benchmarks;
  - theorems showing that mutations preserve the linked-leaf invariant.

  ### Scope

  This PR implements the SHA-256 variant. Generic hash backends and a Poseidon
  variant remain possible follow-ups.

  The package serves Ethereum and ZK protocol research. Fulu and Gloas consensus
  transition behaviour remains unchanged.

  ## LICENSE (LGPL version 3)


leolara:
@protocolwhisper I have not reviewed this yet, but it is interesting.

One thing I would note is that we can very easily make it generic on the hash function, like we do with SSZ here: https://github.com/etheorem/etheorem/blob/main/packages/SizzLean/SizzLean/Hasher/Class.lean#L37-L51

I would not import the class from SizzLean, as SizzLean would become a unecasary dependency, but we can for the type being having a similar class here.

As we have implemented Poseidon2 here: https://github.com/etheorem/etheorem/tree/main/packages/LeanPoseidon

It would work with Sha256, FFI Sha256 and Poseidon2.

leolara:
Also, in another PR we could implement the ZK circuit with this: https://github.com/Verified-zkEVM/clean/

protocolwhisper:
That doesn't bad at all def something I'll like to explore, yes let me move this to the correct structure

leolara:
Yes, of course, the ZK circuit should be a different PR, but let's abstract the hash function, I don't think there is a reason to don't abstract it, right?

Also, look in CLAUDE.md and https://github.com/etheorem/etheorem/blob/main/docs/CODING_STYLE.md for coding conventions.

protocolwhisper:
Asking for a review :)

leolara:
Great work on this. The port is faithful to `vplasencia/leanimt-plus` in the places that matter: `recomputeFrom` matches `walkPath`, the compressed `leafIndex` packing matches `_buildProof`, and `Nat.log2 (n-1) + 1` agrees with `Math.ceil(Math.log2(size))`. I checked the encoding end to end against circomlib's `Sha256_2` (27-byte big-endian inputs, one padded block, low 216 bits of the digest) and it lines up exactly. The fixture is a real cross-check rather than a self-consistency check: building the tree from scratch in Lean reproduces the three reference siblings and `leafIndex = 2` from `input-sha256.json`. A 161-operation randomized run over insert / remove / update held every invariant. The one that matters most: no existing leaf ever admits a forged non-membership proof for a value that is in the tree. The `Hasher` seam is the right shape.

Things to fix before merge.

**Build**

1. `lean_lib LeanImtPlus` declares no `globs`, so it covers only `LeanImtPlus.lean` and its transitive imports. Delete `Hasher/Poseidon2.olean` and `Hasher/Sha256Ffi.olean`, run `lake build LeanImtPlus`, and it reports success without rebuilding either. Two of the three shipped adapters are never typechecked by `just build`. Add `globs = ["LeanImtPlus.*"]`.
2. `packages/LeanImtPlus/lakefile.lean` has no C target, no code generation, and no dynamic git target, so CLAUDE.md's Don'ts put it in `lakefile.toml`. `LeanSha256/lakefile.toml` and `EthCLSpecs/lakefile.toml` already express every construct used here (`licenseFiles`, `moreLeancArgs`, `[[require]] path`, `defaultTargets`, `roots` / `globs`). The same edit lands fix 1.
3. The `LeanHazmatSha256/lakefile.lean` aarch64 fallback change is unrelated to LeanIMT+. Please split it into its own PR.

**Code**

4. `Tree.generateProof` builds the level array twice. `tree.root` computes `levels tree` and keeps one node, then the next line recomputes the whole thing. That doubles the hashing on every proof. Compute `treeLevels` first and read the root out of it.
5. `Sha256.max216` has no references anywhere.
6. `Sha256Spec` and `Sha256Ffi` differ only in `compress`; the other seven fields are identical. A `Sha256.hasherFor (hash : ByteArray → ByteArray)` helper collapses both.
7. `Tree.leaves`, `Tree.indexOf?`, and `Tree.size` are `Id.run do` scans where `filter`, `findIdx?`, and a direct count read shorter. `indexOf?` also scans past the match with no early exit, and it sits on the hot path of insert, remove, update, and proof generation.

**Conventions**

8. The library has no `example` blocks (SizzLean has 29, LeanPoseidon 6, LeanSha256 3). The README's `membershipExample` is a fenced block that nothing compiles, so it can rot silently. Move it into `LeanImtPlus.lean`.
9. `(H := H)` appears 17 times and `@[reducible]` sits on seven instances, neither annotated. Both are load-bearing: drop `@[reducible]` and the `(specRoot : Nat)` and `(specProof.siblings : Array Nat)` ascriptions in `Backends/Sha256Ffi.lean` stop elaborating. CLAUDE.md names both idioms as ones to explain at first use.
10. The four `native_decide` examples in `Fixture.lean` carry no docstring, unlike the ones in `Tree.lean` and the two backend suites. CLAUDE.md asks for the compiler-trust commitment to be visible, and its case 4 prefers kernel `decide` for `Sha256Spec` goals. A single `Sha256.hash2` does reduce in the kernel, so one line on why whole-tree goals use `native_decide` would settle it.

**Docs**

11. About ten "standalone island" and "nothing depends on it" claims about LeanPoseidon are now stale. Among them: `CLAUDE.md:161`, the diagram at `docs/monorepo-arch.md:15`, `packages/LeanPoseidon/LeanPoseidon.lean:23`, and the precompile rationale at `packages/LeanPoseidon/lakefile.lean:46`, which justifies skipping precompilation with "no consumers".
12. The README covers the AVL to linear-scan swap. Worth also noting that Merkle levels are rebuilt from scratch on every `root` and `generateProof` call, where upstream `_recompute` touches O(log n) nodes per mutation.

Only 1 and 2 are blocking from my side. The rest are cheap follow-ups.

PR #25, opened 2026-07-17
feat(gloas-builder-pending-payments-proof): prove processBuilderPendingPayments's withdrawal and payment-window behavior

Mouzayan:
## Summary

Adds `EthCLSpecs/Proofs/BuilderPendingPayments.lean`, proving the local
before/after behavior of one `processBuilderPendingPayments` call:

- Qualifying previous-epoch withdrawals are processed in slot order, preserving
  order and duplicates.
- The unconditional result models `SSZList.push`'s capacity clamp exactly.
- Under an explicit
  `size + qualifying.length ≤ builderPendingWithdrawalsLimit` hypothesis, every
  qualifying withdrawal is appended.
- The payment window shifts down by `SLOTS_PER_EPOCH`: the old upper half moves
  into the lower half and the new upper half is empty.

`processBuilderPendingPayments_run` ties these results directly to the actual
function execution and characterizes the two modified fields through `sszGet`.
`processBuilderPendingPayments_run_of_fits` provides the capacity-guarded
full-append corollary.

## Supporting Changes

- Adds generic lemmas characterizing repeated bounded `SSZList.push`.
- Proves the stateful withdrawal loop reduces to an ordered pure fold.
- Adds lemmas characterizing both regions of the shifted payment window.
- Updates `CONSENSUS_PROOF_CANDIDATES.md` and marks the candidate **In review**.
- Updates the Proofs section of `IMPLEMENTATION_NOTES.md`.

## Notes for Reviewers

- This is a one-call accounting proof. It does not establish protocol-wide
  exactly-once settlement or reason about interactions with
  `settleBuilderPayment` or `processProposerSlashing`.
- No capacity-headroom invariant is assumed or proved. The unconditional theorem
  models clamping; the full-append corollary requires explicit headroom.
- Conclusions use `sszGet` rather than raw `State` equality because cached states
  may have different pending-write representations while exposing the same field
  values.
- The proofs use induction, rewriting, and `omega`; no mathlib,
  `native_decide`, `bv_decide`, `sorry`, or new axioms are introduced.

## Test Plan

- [x] `lake build`
- [x] `lake build EthCLSpecs`
- [x] `just ethcl-test`
- [x] `just lint`
- [x] `just test`


packages/EthCLSpecs/EthCLSpecs/Proofs/BuilderPendingPayments.lean
```diff
@@ -0,0 +1,310 @@
...
+open EthCLSpecs.Fulu (Preset Gwei getTotalActiveBalance)
+open EthCLSpecs.Fulu.Const (slotsPerEpoch builderPaymentThresholdNumerator
+  builderPaymentThresholdDenominator builderPendingWithdrawalsLimit)
+
+/-- Folding `SSZList.push` over a list of values lands on the original array
+plus however much of the values list fits under `cap`: once the list is at
+capacity, `SSZList.push`'s own `if` makes every further push a no-op, so only
+the first `cap - xs.val.size` values appear in the result, in order. No
+additional capacity hypothesis is required. -/
+theorem sszListFoldlPush_val {α : Type} {cap : Nat} (xs : SSZList α cap) (vs : List α) :
```
leolara: Perhaps this proof belongs to the SizzLean package and can be reused here

leolara (approved):
I approve this with a small nit pick:  getTotalActiveBalance should be used from Gloas, not the Fulu one. In the end everything will be relaborated in Gloas so there are no direct dependencies between them that don't go throught the framework.

leolara:
Sorry, after I merged your other PR it needs some conflict solving

Mouzayan:
> Sorry, after I merged your other PR it needs some conflict solving

Thanks, Leo.  Addressed the feedback:
- Resolved merge conflicts.
- Removed the unused `getTotalActiveBalance` Fulu import.
- Moved the `SSZList.push` proofs into the SizzLean package and reused them in the consensus proof.

Ready for another look whenever you have time.

Issue #27, opened 2026-07-20
Mixed-field containers: extend the three central theorems (containerVar)


irajgill:
## Motivation

`SSZType.Supported` / `SSZType.BasicSupported` only cover `.container fs` when every field is fixed-size (`containerFixed`, gated on `allFixedSize fs = true`). Containers with at least one variable-size field (a `list` or `bitlist` field) are outside `Supported` entirely, even though the codec fully implements them:

- `Spec/Serialize.lean`'s `serializeFieldsAux` builds the offset-table wire format for mixed field lists.
- `Spec/Deserialize.lean`'s `.container` arm has a complete non-`allFixedSize` branch (`extractFieldOffsets` + `deserializeVarFields`) that decodes them back.

This is the flagship remaining SSZ proof gap, flagged in the README and `docs/PLAN.md` Stage 18 table ("mixed-field `.container` (≥1 variable-size field) | ⏸ outside `Supported` | needs `Supported` extension first").

## Scope

Add a `containerVar` constructor to `SSZType.Supported`, `SSZType.SupportedBounded`, and `SSZType.BasicSupported` (alongside the existing `containerFixed`, not replacing it), and prove `decode_encode`, `serialize_injective`, `encode_size_le_max` for it. `serialize_injective` inherits automatically once `decode_encode` closes.

Preconditions on the new `BasicSupported.containerVar fs` constructor:
- `BasicSupportedFields fs` (new pointwise predicate: `BasicSupported` per field, without the `isFixedSize` constraint `BasicSupportedFieldsFixed` carries).
- `allFixedSize fs = false` (routes the decoder to the offset-table branch).
- `maxByteLengthFields fs < MAX_LENGTH` (the uint32-offset-overflow guard: the encoder writes `Nat.toUInt32 varOff`, so every running offset must stay under `2^32`, which the schema-derived `maxByteLengthFields` bounds).

## Approach

The core of the proof is a roundtrip walker over `deserializeVarFields`, mirroring `decode_encode_containerFixed_aux`'s mutual-recursion shape but threading a stronger invariant: absolute buffer offsets that tile the variable-body region with the encoder's per-field bodies, recovered from the pre-extracted offset list. Staged as two PRs:
1. Groundwork lemmas (encoder-size accounting, the uint32 offset codec bridge, an offset-extraction inverse) with no predicate changes.
2. The predicates, the walker, the three dispatcher arms, acceptance gates, and docs.

[...]

PR #18, opened 2026-07-14
feat(sizzlean): close the uintN 128/256 arms of the central theorems

irajgill:
## Summary

Closes #17. Extends the three central theorems (`decode_encode`, `serialize_injective`, `encode_size_le_max`) to the two remaining SSZ basic integer widths, **`uintN 128`** and **`uintN 256`**, filling the last gap in the basic-integer row of the proof-coverage table. `uint256` backs `ExecutionPayload.base_fee_per_gas` (Bellatrix onward); `uint128` appears in several spec containers.

Follow-up to #10 (bit shapes).

## Approach: a different, axiom-cleaner proof than the narrow arms

The `uintN 8/16/32/64` arms (`Proofs/UInt.lean`) serialize through fully-unrolled fixed-width LE writers/readers and close by `bv_decide`, which adds one `Lean.ofReduceBool` SAT-certificate axiom per arm.

The wide widths take the `Nat`-based `natToLEBytes` (encode) / `readNatLE` (decode) codec, which recurses on the width rather than unrolling. So the natural proof is a `Nat`-digit induction on the little-endian expansion:

```
readNatLE (natToLEBytes w n .empty) 0 w = some (n % 256 ^ w)
```

The digit step is `Nat.mod_mul` (`x % (a·b) = x % a + a·(x/a % b)`). This uses **no `bv_decide` / `native_decide`**, so the wide arms add **no axioms** beyond the standard kernel three (`propext`, `Classical.choice`, `Quot.sound`), verified by `#print axioms`:

```
'decode_encode_uintN128' depends on axioms: [propext, Classical.choice, Quot.sound]
'decode_encode_uintN256' depends on axioms: [propext, Classical.choice, Quot.sound]
'encode_size_le_max_uintN128' depends on axioms: [propext, Quot.sound]
```

## Changes

- **New `Proofs/UIntWide.lean`**: the codec-inverse lemma chain (`size_natToLEBytes`, `get!_natToLEBytes`, `readNatLEAux_value`, `readNatLE_natToLEBytes`) plus the four arm lemmas.
- **`Spec/BasicSupported.lean`**: new `uintN128` / `uintN256` constructors.
- **Dispatchers**: `decode_encode` (`Roundtrip.lean`), `encode_size_le_max` (`SizeBound.lean`), and `size_serialize_eq_fixedByteSize` (`SerializeSize.lean`) gain the two arms; `serialize_injective` inherits automatically.
[...]

PR #28, opened 2026-07-20
feat(sizzlean): groundwork for mixed-field container proofs

irajgill:
## Summary

First of two PRs closing #27 (mixed-field `.container` proofs, the flagship remaining gap in `SSZType.Supported`). This PR is groundwork only: a new `Proofs/ContainerVar.lean` with the codec-level lemmas the roundtrip proof needs. No predicate or dispatcher changes, so it builds and reviews standalone.

- **uint32 offset codec bridge** (`readUInt32LE_uint32LE_append`, `readUInt32LE_append_shift`, `toNat_toUInt32_of_lt`): the `uint32LE` / `readUInt32LE` round-trip, plus the shift needed to read an offset placeholder embedded partway through a buffer.
- **Extract-middle** (`extract_middle`): slicing the middle piece out of a three-way `ByteArray` append, for identifying a variable field's body via `b.extract curOff nextOff` in the follow-up roundtrip walker.
- **Field-list plumbing devices** (`FieldsFixedSizeOk`, `FieldsMaxSizeOk`, `varOffsetsOf`): local, non-exported structural `def`s that thread per-field correctness facts alongside a heterogeneous field list, ahead of `BasicSupportedFields` (the real, proof-carrying predicate) landing in the follow-up PR.
- **Encoder accounting** (`size_serializeFieldsAux_fixedSection`) and the **size walker** (`size_serializeFieldsAux_le_maxByteLengthFields`): `serializeFieldsAux`'s output sizes match the schema-derived bounds.
- **Offset-extraction inverse** (`extractFieldOffsets_serializeFieldsAux`): reading `extractFieldOffsets` back off the encoder's own output recovers exactly the running offsets it wrote. The hard direction; its docstring explains why the invariant doesn't need to track where the variable region physically sits.

Also promotes `extractFieldOffsets` from `private` to `protected` in `Spec/Deserialize.lean` (was unreachable from proof files entirely), mirroring the `natToLEBytes` / `readNatLE` convention from #18.

Stacks on #21 (same `Spec/BasicSupported.lean` neighborhood is untouched here, but the branch is cut from `fix/supported-uintN-wide` per the scoping issue's plan; the diff below will include #21's commit until it merges).

## Verification

- `lake build SizzLean SizzLeanTests`: green.
- Axiom audit (`#print axioms`) on every new theorem: only `propext`, `Classical.choice`, `Quot.sound`, plus one `bv_decide` axiom on the two theorems that touch the uint32 codec bridge, same trust class as the narrow `uintN` arms.

## Test plan

- [x] `lake build SizzLean SizzLeanTests` green
- [x] Axiom audit on the five new theorem groups

### DIFF
--- a/packages/SizzLean/SizzLean/Proofs/ContainerVar.lean
+++ b/packages/SizzLean/SizzLean/Proofs/ContainerVar.lean
@@ -1,0 +1,433 @@
+import SizzLean.Spec.Serialize
+import SizzLean.Spec.Deserialize
+import SizzLean.Spec.MaxByteLength
+import Std.Tactic.BVDecide
+
+/-!
+# `SizzLean.Proofs.ContainerVar`: groundwork for mixed-field `.container fs`
+
+Mixed-field containers (at least one variable-size field, `list` /
+`bitlist`) are the flagship remaining gap in `SSZType.Supported`:
+the codec ([`Spec/Serialize.lean`](../Spec/Serialize.lean)'s
+`serializeFieldsAux`, [`Spec/Deserialize.lean`](../Spec/Deserialize.lean)'s
+non-`allFixedSize` `.container` branch) fully implements the
+offset-table wire format, but no `BasicSupported` constructor
+claims it yet, and no theorem closes it. This module lays the
+groundwork; the predicates (`containerVar` on `BasicSupported` /
+`Supported` / `SupportedBounded`) and the roundtrip walker land in
+a follow-up, once this file's lemmas are available to build on.
+
+## Wire format (what the lemmas below characterize)
+
+```
+fixed section                         variable section
+┌────────┬──────────┬────────┬──────┐ ┌────────┬────────┐
+│ A bytes│ offset(B)│ C bytes│offset│ │ B body │ D body │
+│(fixed) │ (uint32) │(fixed) │ (D)  │ │        │        │
+└────────┴──────────┴────────┴──────┘ └────────┴────────┘
+```
+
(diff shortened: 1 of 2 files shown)


packages/SizzLean/SizzLean/Proofs/ContainerVar.lean
```diff
@@ -0,0 +1,433 @@
...
+placeholder to `.1` and their own body to the `.2` accumulator,
+then advance `varOff` by the body's size. The decoder's
+`extractFieldOffsets` walks `fs` against the *same* running `off`
+to read the placeholders back out, before `deserializeVarFields`
+uses them to slice each variable field's body out of the variable
+region.
+
+## Lemma path
+
+The five groundwork facts, in the order they compose:
```
leolara: The intro here says "The five groundwork facts", but six numbered items follow. Item 3 (`FieldsFixedSizeOk` / `FieldsMaxSizeOk` / `varOffsetsOf`) are the plumbing `def`s, not proven facts. Either count six, or drop the devices from the "facts" framing, for example "The groundwork, in the order it composes:".

leolara (approved):

packages/SizzLean/SizzLean/Proofs/ContainerVar.lean
irajgill: Fixed, dropped the count and the "facts" framing per your suggestion: "The groundwork, in the order it composes (proven facts, plus the plumbing `def`s of item 3 that thread per-field data through them):". Pushed on both this branch and the stacked #29.

PR #29, opened 2026-07-20
feat(sizzlean): close the remaining central-theorem arms for mixed-field containers

irajgill:
## Summary

Second of two PRs closing #27 (mixed-field `.container` proofs). Where #28 landed the codec-level groundwork, this PR wires up the predicates and the two proof arms that actually close the three central theorems for containers that mix fixed- and variable-size fields.

- **`BasicSupportedFields` / `SupportedFields` / `SupportedBoundedFields`** (`Spec/BasicSupported.lean`, `Spec/Supported.lean`): pointwise field-list predicates like their `*FieldsFixed` cousins, but with no `isFixedSize` constraint, since the offset-table decode path handles fixed and variable fields alike.
- **`containerVar` constructor** on all three of `BasicSupported` / `Supported` / `SupportedBounded`. On `BasicSupported` it also carries `allFixedSize fs = false` (routes decode to the offset branch) and `maxByteLengthFields fs < MAX_LENGTH` (the uint32-offset-overflow guard, see the constructor's docstring). `supported_of_basicSupported` (the machine-checked subset theorem from #21) grows a matching arm, so the two predicates can't drift apart on this constructor either.
- **`decode_encode_containerVar_aux`** (`Proofs/Roundtrip.lean`'s mutual block, alongside `decode_encode`): the offset-table roundtrip walker. Threads two `ByteArray.extract` invariants pinning the remaining fixed prefix and variable region to `serializeFieldsAux`'s output, decomposed at each cons step via #28's `extract_split`; new helpers `varOffsetsOf_head_getD` and `allFixedSize_of_varOffsetsOf_eq_nil` land in `Proofs/ContainerVar.lean` alongside it.
- **Dispatcher wiring**: `containerVar` arms on `decode_encode` (above), `encode_size_le_max` (`SizeBound.lean`, via the size walker from #28), and an absurd arm on `SerializeSize.lean`'s `size_serialize_eq_fixedByteSize` (a `containerVar` container is never fixed-size).
- **Acceptance gates** (`SizzLeanTests/ReprExamples.lean`): four `containerVar` shapes, variable field last, variable field first, two variable fields (offset chaining across bodies), and a concrete empty-list value exercising the `curOff = nextOff = bufEnd` edge.
- **Docs**: README (top-level bullet, proof-coverage table, per-field-type table, track-in-progress), `docs/ARCHITECTURE.md` (proof-status paragraph and the Phase 5 roadmap row), `docs/PLAN.md` (Stage 18 section and its Phase-5 summary row).

Stacks on #28 (itself stacked on #21); the diff below will include both until they merge.

## Verification

- `lake build SizzLean SizzLeanTests`: green, no warnings.
- Axiom audit (`#print axioms`) on the three central theorems:
  - `decode_encode` / `serialize_injective`: `propext`, `Classical.choice`, `Quot.sound`, plus four `bv_decide` axioms (the three narrow `uintN16/32/64` arms, unchanged, plus one new axiom from the mixed-field container arm's uint32 offset-codec bridge).
  - `encode_size_le_max`: `propext`, `Quot.sound` only, no `bv_decide`.

## Test plan

- [x] `lake build SizzLean SizzLeanTests` green
- [x] Axiom audit on `decode_encode`, `serialize_injective`, `encode_size_le_max`
- [x] Four new acceptance gates compile


leolara:
Review of `8896968`. The proof engineering holds up. I rebuilt locally and audited the axioms independently, and both match the description. What needs work is the branch state and the surrounding docs.

## Confirmed good

- `lake build SizzLean SizzLeanTests` green locally; all seven CI checks green.
- Axiom audit reproduces exactly. `decode_encode` / `serialize_injective` carry `propext`, `Classical.choice`, `Quot.sound`, the three `uintN16/32/64` bv_decide axioms, and `readUInt32LE_uint32LE_append._native.bv_decide.ax_1_14`. `encode_size_le_max` carries `propext` and `Quot.sound`.
- No runtime code changes. Every changed `.lean` is a `Prop` inductive, a proof, or a test, so there is no conformance risk.
- Exhaustiveness on the three `BasicSupported` matchers is compiler-enforced, and nothing eliminates `Supported` or `SupportedBounded`, so the new constructors break no existing match.
- The offset arithmetic chains correctly: `fixedSectionSizeFields fs + varSection ≤ maxByteLengthFields fs < 2^32` is exactly the bound `extractFieldOffsets_serializeFieldsAux` needs.

## Blockers

### 1. Rebase onto main

`gh pr view` reports `CONFLICTING / DIRTY`. #28 landed as squash `230726b`, so `53b867e` and `7011f5a` are duplicate content now. After the rebase the diff should be the 11 files this PR owns.

### 2. The coverage claim is wrong; `maxByteLengthFields fs < MAX_LENGTH` excludes real containers

`README.md:64` and `docs/ARCHITECTURE.md:132` both say the theorems are proved for every shape *except* `vector` / `list` over a variable-size element type. `containerVar` also requires `maxByteLengthFields fs < MAX_LENGTH`, a schema-level bound that rules out the containers this library targets. Machine-checked with the `balances` shape:

```lean
example : ¬ (SSZType.maxByteLengthFields
    [SSZType.uintN 64, SSZType.list (.uintN 64) (2 ^ 40)] < MAX_LENGTH) := by
  native_decide   -- succeeds
```

`List[uint64, VALIDATOR_REGISTRY_LIMIT]` is `2^43` bytes of static max on its own, so every `BeaconState` and `BeaconBlockBody` stays outside `BasicSupported`. The guard itself is right, offsets are `uint32`, and a value-level replacement would change all three theorem statements. So keep the design and fix the claim:

- `README.md:64` and `ARCHITECTURE.md:132`: name the second exception.
- `README.md:210`: the Preconditions column reads "offset-table codec" while every other row spells its preconditions out. Add `maxByteLengthFields fs < MAX_LENGTH`.
- `README.md:270`: the "In one line" sentence needs the same qualifier.

Worth filing a follow-up issue for the value-level relaxation, since it is what stands between this arm and `BeaconState`.

### 3. Two errors in the proof-coverage table

- `README.md:212`, footnote ¹ still says "exactly these three axioms". It is four now, and footnote ⁶ supplies the fourth. Footnote ¹ is the line a reader trusts for the total.
- `README.md:210` marks ⁶ on all three columns including `encode_size_le_max`, and ⁶ describes a bv_decide axiom. The audit shows `encode_size_le_max` carries `propext` and `Quot.sound` only. Rows ¹ and ⁵ already set the convention: mark the two roundtrip columns and leave the size column bare.

## Should fix before merge

### 4. `Proofs/ContainerVar.lean:9-18` is now false

The intro still reads "no `BasicSupported` constructor claims it yet, and no theorem closes it. This module lays the groundwork; the predicates … and the roundtrip walker land in a follow-up." This PR is that follow-up. The section header at `:442` and the new lemma-path item 7 also narrate how the file grew ("added once the roundtrip walker … needed them"); both should read as statements about what the file contains today.

### 5. `Repr/Class.lean:55-61` was missed by the sweep

It still tells users that a shape with "mixed fixed/variable-size container fields" gets total serialize / deserialize but no verified roundtrip. This is the user-facing explanation of the `SSZ.roundtrip` gate, so it is the highest-traffic stale spot in the tree.

### 6. `Proofs/Roundtrip.lean`'s docstring is out of step with `SizeBound.lean`'s

`SizeBound.lean:22-29` correctly became "three-way mutual block". Roundtrip's heading at `:45` still reads "The mutual `decode_encode` / `decode_encode_containerFixed_aux`" and its prose describes two members. The arm-to-file table at `:22-29` never gained a `ContainerVar.lean` row either.

### 7. Document the new dependency between two central theorems

Roundtrip.lean now imports `SizzLean.Proofs.SizeBound`, and the `containerVar` arm calls `encode_size_le_max_containerVarFields_aux`, so `decode_encode` transitively depends on `encode_size_le_max`. That is inherent, since bounding the encoded size is how the offsets are shown not to overflow, and the audit confirms it costs no axioms. ARCHITECTURE presents the three theorems as independent, so one sentence in Roundtrip's module docstring closes the gap.

## Minor

8. `Spec/Supported.lean:94` says "The `Bool` argument mirrors `allFixedSize fs`". There is no `Bool` argument; it is a proof of `allFixedSize fs = false`.
9. `Proofs/Roundtrip.lean:144` references "(`extractFieldOffsets_serializeFieldsAux`, PR 1)", which dangles once this merges.
10. `README.md:258`, the nested-container row reads "with `BasicSupportedFieldsFixed fs'` or `BasicSupportedFields fs'`". `BasicSupportedFields fs'` alone is not a witness for `BasicSupported (.container fs')`; the constructor takes two more arguments. In that case the nested container is itself variable-size, so the row also wants the ᵛ marker.
11. `SupportedBounded.containerVar` is reachable by no proof and guarded by no subset theorem. `supported_of_basicSupported` machine-checks `BasicSupported → Supported`, and nothing links `SupportedBounded` to either. This PR makes it three hand-maintained parallel families with a single checked edge, so a `supported_of_supportedBounded` companion or a tracking note would help.
12. The `emptyListContainer` gate (`ReprExamples.lean:295-310`) instantiates a universally quantified theorem, so it cannot fail independently of the first gate's shape. Its docstring says the decoder guard "must accept `curOff = nextOff = bufEnd`", which reads like a test of that guard; what actually discharges it is `decode_encode_containerVar_aux`'s `h_guard` step. Reword so the gate does not imply coverage it does not add.

## Observation

`deserializeVarFields`'s variable branch discards the consumed-byte count (`Deserialize.lean:532`) where the fixed branch guards with `if used ≠ sz then .error .trailingBytes`, and the top-level container arm returns `.ok (v, b.size)` without checking that the offsets tile the buffer. The `containerVar` arm is correct as stated, since `serialize_injective` is "distinct values, distinct encodings". It establishes nothing about the decoder rejecting non-canonical offset tables, and the README marks non-malleability ✅ for the row. A sentence separating the two properties, or an issue tracking decoder canonicity, would keep the claim from being read wider than it is. The `ssz_generic containers/invalid` vectors are the current cover.

irajgill:
Addressed the review on `8896968` (now rebased as `aee72b3` + docs `ae221a5`).

**Blockers**
1. Rebased onto current `main`; #28 squash duplicates dropped. Diff is the 11 files this PR owns (`Repr/Class.lean` included).
2. Named `maxByteLengthFields fs < MAX_LENGTH` as a second exception in README / ARCHITECTURE / PLAN / BasicSupported. Opened #61 for the value-level relaxation.
3. Footnote ¹ now says four axioms; ⁶ is only on the two roundtrip columns.

**Should-fix**
4. `ContainerVar.lean` intro and lemma-path item 7 rewritten in present tense.
5. `Repr/Class.lean` updated: `containerVar` is inside the gate; outside examples are variable-element `vector`/`list` and oversize schemas.
6. Roundtrip docstring is now a three-way mutual block, with a `ContainerVar.lean` row in the arm table.
7. Added a "Dependence on `encode_size_le_max`" section to Roundtrip's module docstring.

**Minor**
8–12 all addressed. Also separated injectivity from decoder canonicity in the README, pointing at the `ssz_generic containers/invalid` vectors.

Build and axiom audit still match what you reported on the previous tip.

leolara:
Great work on this. Every point landed, and the re-check at `ae221a5` came
back clean across the board.

On 11: the note is one of the two things I asked for, so this is fine as it
stands. If a fourth parallel family ever shows up,
`supported_of_supportedBounded` becomes the durable answer.

## Three cosmetics, none blocking

1. `docs/ARCHITECTURE.md:139` has a 109-character line left by the reflow
   ("…until those close too. The asterisk on…") in a file that wraps near 70.
2. Roundtrip's arm-to-file table gained a `.containerVar fs` row but still has
   no `.containerFixed fs` → `Proofs/ContainerFixed.lean` row, while the prose
   just below says that file ships the substantive helpers. Worth adding for
   symmetry.
3. The status blockquote at `README.md:8` says the cut "now covers mixed-field
   containers too" with no qualifier. The detailed bullet 50 lines down carries
   the caveat, so this is only about the line a skimmer reads first.

Fix those in this PR or a follow-up, either works. Approving.

irajgill:
Fixed the three cosmetics on top of `ae221a5`:

1. Wrapped the long ARCHITECTURE proof-status line.
2. Added `.containerFixed fs` → `Proofs/ContainerFixed.lean` to Roundtrip's arm table.
3. Qualified the README status blockquote with the `MAX_LENGTH` schema guard.

Thanks for the re-check and the approve.

PR #30, opened 2026-07-21
feat(gloas-update-checkpoints-proof): prove updateCheckpoints checkpoint monotonicity

Mouzayan:
## Summary

Adds `EthCLSpecs/Proofs/UpdateCheckpoints.lean`, proving monotonicity for
`EthCLSpecs.Gloas.updateCheckpoints`:

- Each justified/finalized checkpoint either remains unchanged when its
  candidate is not newer, or becomes exactly the candidate when its epoch is
  strictly greater. Both outcomes include the guard condition that selected them.
- Two corollaries prove that `updateCheckpoints` never lowers either recorded epoch.
- The proof is scoped to the Gloas declaration. Fulu's separate declaration is
  out of scope.

## Supporting Changes

- Marks `updateCheckpoints` **In review** in
  `CONSENSUS_PROOF_CANDIDATES.md`.
- Adds the proof module to `IMPLEMENTATION_NOTES.md`.
- Updates `Proofs.lean`'s module documentation to include ordinary case analysis.

## Notes for Reviewers

- This PR proves monotonicity for each invocation of `updateCheckpoints`; it
  does not establish a protocol-wide invariant over arbitrary fork-choice
  transition histories. Such a theorem would additionally need to cover Store
  initialization and every path that can write either checkpoint field.
- `getForkchoiceStore` writes both fields during genesis/anchor initialization
  and is outside this claim.
- The proofs use `by_cases`, `simp`, and core `UInt64` ordering lemmas. They
  introduce no mathlib dependency, decision procedure, `sorry`, or new axiom.
- `#print axioms` reports only `propext`, `Classical.choice`, and `Quot.sound`.

## Test Plan

- [x] `lake build`
- [x] `lake build EthCLSpecs.Proofs.UpdateCheckpoints`
- [x] `just ethcl-test`
- [x] `just lint`

### DIFF
--- a/packages/EthCLSpecs/EthCLSpecs/Proofs.lean
+++ b/packages/EthCLSpecs/EthCLSpecs/Proofs.lean
@@ -7,8 +7,9 @@ import EthCLSpecs.Proofs.UpdateCheckpoints
 
 Mathlib-free proofs about `EthCLSpecs` declarations, colocated with the specs
 the way `SizzLean.Proofs` is colocated with `SizzLean`: same package, same
-build, `bv_decide` / `decide` / `native_decide` over the spec's own types, no
-mathlib. A theorem that turns out to need mathlib moves to the standalone
+build, closed by whichever of `bv_decide`, `decide`, `native_decide`, or plain
+case analysis the goal needs, over the spec's own types, no mathlib. A
+theorem that turns out to need mathlib moves to the standalone
 `EthCLProofs` package instead (`docs/SPECS_ARCHITECTURE.md` §11), the
 `LeanPoseidonProofs` containment pattern, so mathlib never reaches this
 library, the framework, the runner, or the conformance path.
--- a/packages/EthCLSpecs/EthCLSpecs/Proofs.lean
+++ b/packages/EthCLSpecs/EthCLSpecs/Proofs.lean
@@ -1,5 +1,6 @@
 import EthCLSpecs.Proofs.BuilderIndex
 import EthCLSpecs.Proofs.InitializePtcWindow
+import EthCLSpecs.Proofs.UpdateCheckpoints
 
 /-!
 # `EthCLSpecs.Proofs`: consensus-spec theorems (index)
@@ -18,4 +19,6 @@ Re-exports:
   (`isBuilderIndex`, `toBuilderIndex`, `convertBuilderIndexToValidatorIndex`).
 * `EthCLSpecs.Proofs.InitializePtcWindow`: the seeded `ptcWindow`'s two
   regions (`initializePtcWindow`).
+* `EthCLSpecs.Proofs.UpdateCheckpoints`: `Gloas.updateCheckpoints` checkpoint
+  monotonicity, the justified/finalized epoch never decreases.
 -/
--- a/packages/EthCLSpecs/EthCLSpecs/Proofs.lean
+++ b/packages/EthCLSpecs/EthCLSpecs/Proofs.lean
@@ -7,12 +7,12 @@ import EthCLSpecs.Proofs.UpdateCheckpoints
 
 Mathlib-free proofs about `EthCLSpecs` declarations, colocated with the specs
 the way `SizzLean.Proofs` is colocated with `SizzLean`: same package, same
-build, closed by whichever of `bv_decide`, `decide`, `native_decide`, or plain
-case analysis the goal needs, over the spec's own types, no mathlib. A
-theorem that turns out to need mathlib moves to the standalone
-`EthCLProofs` package instead (`docs/SPECS_ARCHITECTURE.md` §11), the
-`LeanPoseidonProofs` containment pattern, so mathlib never reaches this
-library, the framework, the runner, or the conformance path.
+build. Each theorem is closed by whichever tactic its goal needs, `bv_decide`,
+`decide`, `native_decide`, or plain case analysis. Always over the spec's own
+types, never mathlib. A theorem that turns out to need mathlib moves to the standalone
+`EthCLProofs` package instead (`docs/SPECS_ARCHITECTURE.md` §11), following the
+`LeanPoseidonProofs` containment pattern. Mathlib never reaches this library,
+the framework, the runner, or the conformance path.
 
 Re-exports:
 
--- a/packages/EthCLSpecs/EthCLSpecs/Proofs.lean
+++ b/packages/EthCLSpecs/EthCLSpecs/Proofs.lean
@@ -21,5 +21,6 @@ Re-exports:
 * `EthCLSpecs.Proofs.InitializePtcWindow`: the seeded `ptcWindow`'s two
   regions (`initializePtcWindow`).
 * `EthCLSpecs.Proofs.UpdateCheckpoints`: `Gloas.updateCheckpoints` checkpoint
-  monotonicity, the justified/finalized epoch never decreases.
+  monotonicity, the justified/finalized epoch never decreases. Its theorems sit
+  in `EthCLSpecs.Proofs.Gloas`, since `updateCheckpoints` exists in both forks.
 -/
--- a/packages/EthCLSpecs/EthCLSpecs/Proofs/UpdateCheckpoints.lean
+++ b/packages/EthCLSpecs/EthCLSpecs/Proofs/UpdateCheckpoints.lean
@@ -23,12 +23,10 @@ open EthCLSpecs.Gloas (Store updateCheckpoints Checkpoint)
 open EthCLSpecs.Fulu (Preset)
 open EthCLLib.Spec (MapKind HasherTag)
 
-variable {map : MapKind} [Preset] [HasherTag]
-  (store : Store map) (j f : Checkpoint)
-
 /-- The resulting justified checkpoint is either unchanged because `j` is not newer,
 or exactly `j` because its epoch is strictly greater. -/
-theorem updateCheckpoints_justifiedCheckpoint_eq_or_advances :
+theorem updateCheckpoints_justifiedCheckpoint_eq_or_advances
+    {map : MapKind} [Preset] [HasherTag] (store : Store map) (j f : Checkpoint) :
     ((updateCheckpoints store j f).justifiedCheckpoint = store.justifiedCheckpoint ∧
         j.epoch ≤ store.justifiedCheckpoint.epoch) ∨
     ((updateCheckpoints store j f).justifiedCheckpoint = j ∧
@@ -44,7 +42,8 @@ theorem updateCheckpoints_justifiedCheckpoint_eq_or_advances :
 
 /-- The resulting finalized checkpoint is either unchanged because `f` is not newer,
 or exactly `f` because its epoch is strictly greater. -/
-theorem updateCheckpoints_finalizedCheckpoint_eq_or_advances :
+theorem updateCheckpoints_finalizedCheckpoint_eq_or_advances
+    {map : MapKind} [Preset] [HasherTag] (store : Store map) (j f : Checkpoint) :
     ((updateCheckpoints store j f).finalizedCheckpoint = store.finalizedCheckpoint ∧
         f.epoch ≤ store.finalizedCheckpoint.epoch) ∨
     ((updateCheckpoints store j f).finalizedCheckpoint = f ∧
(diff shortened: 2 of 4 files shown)


leolara:
## Review, plus seven follow-up commits

The proofs are correct and the claims in the description check out. I verified independently rather than taking the checklist on trust:

- `lake build` green, 370 jobs.
- `#print axioms` on all theorems reports `propext`, `Classical.choice`, `Quot.sound` and nothing else, before and after my changes.
- `just lint` passes.
- The module docstring's scope claim holds. Grepping Gloas for writes to `justifiedCheckpoint` / `finalizedCheckpoint` turns up exactly two sites, `updateCheckpoints` at `Gloas/ForkChoice.lean:471-472` and `getForkchoiceStore` at `:815`. The docstring names both and correctly excludes the second.

The subtle step is sound too. The definition's outer guard tests `store'.finalizedCheckpoint.epoch` on the already-updated store while the theorem states it against the original `store`, which works because the first update touches only `justifiedCheckpoint`. Letting `simp` discharge that instead of asserting it in prose was the right call.

I pushed the seven review items as separate commits rather than leaving them as comments. Each is described below.

---

### 1. `f632553` Collapse the duplicated case bullets

Both branch proofs split on the two guards with `<;>` and then discharged four bullets, of which each pair was character-for-character identical. That repetition hid the actual structure: only the outer guard decides the disjunct, and the inner one exists purely so `simp` can reduce the other field's `ite`.

`refine` names it. The outer `by_cases` picks the disjunct and supplies the guard; the inner one closes the equality in both of its cases with one `simp`. Eight bullets become four.

Two things I checked before proposing this, because the original shape looked like it had slack that it does not actually have:

- **The second `by_cases` is load-bearing.** I tried the finalized theorem with only its own guard split and it leaves unsolved goals. Both splits are needed in both theorems.
- **`split` is not the shorter route.** `forkdef` elaborates the definition's `let` into `have store := ...`, and `split` refuses to see through it: `Could not split an 'if' or 'match' expression in the goal`. Both facts are now recorded in the module docstring, see item 5, so nobody rediscovers them the hard way.

### 2. `960ab55` Scope the theorems to a `Gloas` namespace

This is the one item I would call more than polish. `updateCheckpoints` is declared identically in Fulu (`Fulu/ForkChoice.lean:291-293`) and Gloas (`Gloas/ForkChoice.lean:470-472`). With every proof module sharing the flat `EthCLSpecs.Proofs` namespace, a Fulu companion collides on all four names, and by then the rename is a breaking change to whatever depends on them.

The sibling modules do not hit this because `convertBuilderIndexToValidatorIndex` and `initializePtcWindow` are Gloas-only. This module moves to `EthCLSpecs.Proofs.Gloas`, mirroring the namespace the subject sits in and leaving `EthCLSpecs.Proofs.Fulu` free. Cheap now, awkward later.

### 3. `3d94c55` Add the frame condition

The two branch characterizations project one checkpoint field each. Neither says anything about the other seventeen Store fields, so even together they leave open whether `updateCheckpoints` disturbs `time`, `equivocatingIndices`, or any of the maps. The docstring said the file "characterizes both branches exactly"; it characterized the two fields it looked at.

`updateCheckpoints_eq` closes that by rewriting the whole call as a single record update over the original store, so the frame condition falls out of the `{ store with ... }` syntax rather than needing a field-by-field enumeration. Same four-case split, one `simp`. A caller threading a Store through a fork-choice transition needs this, and it costs nine lines.

### 4. `ceac1b1` Hoist the shared binders back to `variable`

`81aa00a` dropped the `variable` block and inlined the same five binders per theorem; the frame condition would have made that a fifth copy. CLAUDE.md's DIP bullet endorses the deleted form verbatim (`variable {α : Type} [SSZ α]` over `(x : SpecificStruct)`), and DRY asks for one canonical home per fact.

I take the point that the sibling modules state binders per theorem, which I assume drove that commit. The cases differ in weight: they carry two binders and no value arguments, this one carries five.

Verified with `#check` that every signature is byte-identical to the inlined form. Lean includes a `variable` only where the statement mentions it, and all five statements mention all five.

### 5. `b424d0a` Lift the shared proof note into the module docstring

The two-line "decide both guards" comment appeared verbatim above each branch proof. CLAUDE.md asks for an idiom to be annotated the first time it appears, not the fifth. It moves to a "shared proof shape" section of the module docstring, where it now also carries the two findings from item 1 that the code cannot show on its own.

### 6. `9b44980` Split the index docstring's overlong sentence

Widening the tactic list in `Proofs.lean` pushed the opening sentence to four commas. CLAUDE.md caps a subphrase at two and asks for a split past that. It becomes three sentences: the colocation, the tactic choice, the mathlib boundary. The sentence after it had the same shape and got the same treatment.

### 7. `38f4f67` Retire the "In review" status marker

Adding this row followed the existing convention, so the PR was consistent with what was there. The convention is the problem, and it is worth killing now rather than growing a fourth instance.

`convertBuilderIndexToValidatorIndex` and `initializePtcWindow` both still read **In review** even though their proofs merged some time ago. Nothing ever flips the marker, so a status that only moves in one direction is recording pull-request history in a surviving artifact. We do not do that anywhere else in the repo, and it is exactly the drift that makes a doc stop being trustworthy.

**We should not have an "In review" state at all.** All three rows now read **Proved** and name their module, including the two merged ones, and the Purpose section states the marker is binary: unmarked while a candidate is open, **Proved** once its module lands. Work in flight is tracked by its pull request, which is already the right place for it and is where this exact conversation lives.

Marking this PR's own row **Proved** before merge is deliberate under that rule. The doc describes the state of the tree it ships in, not the state of the review.

The row also picks up the frame condition from item 3, and the three rows now punctuate consistently.

---

### Not changed

Four lines sit at 101-102 columns. There is no enforced limit and `InitializePtcWindow.lean:47` is already 105, so I left them.

### Verification after all seven

`lake build` green, `just lint` clean, and `#print axioms` on all five theorems still reports only `propext`, `Classical.choice`, `Quot.sound`. No `sorry`, no new axiom, no mathlib.

Happy to drop any of these if you disagree, 2 and 3 are the two I would keep.

leolara (approved):

Mouzayan:
> ## Review, plus seven follow-up commits
> The proofs are correct and the claims in the description check out. I verified independently rather than taking the checklist on trust:
> 
> * `lake build` green, 370 jobs.
> * `#print axioms` on all theorems reports `propext`, `Classical.choice`, `Quot.sound` and nothing else, before and after my changes.
> * `just lint` passes.
> * The module docstring's scope claim holds. Grepping Gloas for writes to `justifiedCheckpoint` / `finalizedCheckpoint` turns up exactly two sites, `updateCheckpoints` at `Gloas/ForkChoice.lean:471-472` and `getForkchoiceStore` at `:815`. The docstring names both and correctly excludes the second.
> 
> The subtle step is sound too. The definition's outer guard tests `store'.finalizedCheckpoint.epoch` on the already-updated store while the theorem states it against the original `store`, which works because the first update touches only `justifiedCheckpoint`. Letting `simp` discharge that instead of asserting it in prose was the right call.
> 
> I pushed the seven review items as separate commits rather than leaving them as comments. Each is described below.
> 
> ### 1. `f632553` Collapse the duplicated case bullets
> Both branch proofs split on the two guards with `<;>` and then discharged four bullets, of which each pair was character-for-character identical. That repetition hid the actual structure: only the outer guard decides the disjunct, and the inner one exists purely so `simp` can reduce the other field's `ite`.
> 
> `refine` names it. The outer `by_cases` picks the disjunct and supplies the guard; the inner one closes the equality in both of its cases with one `simp`. Eight bullets become four.
> 
> Two things I checked before proposing this, because the original shape looked like it had slack that it does not actually have:
> 
> * **The second `by_cases` is load-bearing.** I tried the finalized theorem with only its own guard split and it leaves unsolved goals. Both splits are needed in both theorems.
> * **`split` is not the shorter route.** `forkdef` elaborates the definition's `let` into `have store := ...`, and `split` refuses to see through it: `Could not split an 'if' or 'match' expression in the goal`. Both facts are now recorded in the module docstring, see item 5, so nobody rediscovers them the hard way.
> 
> ### 2. `960ab55` Scope the theorems to a `Gloas` namespace
> This is the one item I would call more than polish. `updateCheckpoints` is declared identically in Fulu (`Fulu/ForkChoice.lean:291-293`) and Gloas (`Gloas/ForkChoice.lean:470-472`). With every proof module sharing the flat `EthCLSpecs.Proofs` namespace, a Fulu companion collides on all four names, and by then the rename is a breaking change to whatever depends on them.
> 
> The sibling modules do not hit this because `convertBuilderIndexToValidatorIndex` and `initializePtcWindow` are Gloas-only. This module moves to `EthCLSpecs.Proofs.Gloas`, mirroring the namespace the subject sits in and leaving `EthCLSpecs.Proofs.Fulu` free. Cheap now, awkward later.
> 
> ### 3. `3d94c55` Add the frame condition
> The two branch characterizations project one checkpoint field each. Neither says anything about the other seventeen Store fields, so even together they leave open whether `updateCheckpoints` disturbs `time`, `equivocatingIndices`, or any of the maps. The docstring said the file "characterizes both branches exactly"; it characterized the two fields it looked at.
> 
> `updateCheckpoints_eq` closes that by rewriting the whole call as a single record update over the original store, so the frame condition falls out of the `{ store with ... }` syntax rather than needing a field-by-field enumeration. Same four-case split, one `simp`. A caller threading a Store through a fork-choice transition needs this, and it costs nine lines.
> 
> ### 4. `ceac1b1` Hoist the shared binders back to `variable`
> `81aa00a` dropped the `variable` block and inlined the same five binders per theorem; the frame condition would have made that a fifth copy. CLAUDE.md's DIP bullet endorses the deleted form verbatim (`variable {α : Type} [SSZ α]` over `(x : SpecificStruct)`), and DRY asks for one canonical home per fact.
> 
> I take the point that the sibling modules state binders per theorem, which I assume drove that commit. The cases differ in weight: they carry two binders and no value arguments, this one carries five.
> 
> Verified with `#check` that every signature is byte-identical to the inlined form. Lean includes a `variable` only where the statement mentions it, and all five statements mention all five.
> 
> ### 5. `b424d0a` Lift the shared proof note into the module docstring
> The two-line "decide both guards" comment appeared verbatim above each branch proof. CLAUDE.md asks for an idiom to be annotated the first time it appears, not the fifth. It moves to a "shared proof shape" section of the module docstring, where it now also carries the two findings from item 1 that the code cannot show on its own.
> 
> ### 6. `9b44980` Split the index docstring's overlong sentence
> Widening the tactic list in `Proofs.lean` pushed the opening sentence to four commas. CLAUDE.md caps a subphrase at two and asks for a split past that. It becomes three sentences: the colocation, the tactic choice, the mathlib boundary. The sentence after it had the same shape and got the same treatment.
> 
> ### 7. `38f4f67` Retire the "In review" status marker
> Adding this row followed the existing convention, so the PR was consistent with what was there. The convention is the problem, and it is worth killing now rather than growing a fourth instance.
> 
> `convertBuilderIndexToValidatorIndex` and `initializePtcWindow` both still read **In review** even though their proofs merged some time ago. Nothing ever flips the marker, so a status that only moves in one direction is recording pull-request history in a surviving artifact. We do not do that anywhere else in the repo, and it is exactly the drift that makes a doc stop being trustworthy.
> 
> **We should not have an "In review" state at all.** All three rows now read **Proved** and name their module, including the two merged ones, and the Purpose section states the marker is binary: unmarked while a candidate is open, **Proved** once its module lands. Work in flight is tracked by its pull request, which is already the right place for it and is where this exact conversation lives.
> 
> Marking this PR's own row **Proved** before merge is deliberate under that rule. The doc describes the state of the tree it ships in, not the state of the review.
> 
> The row also picks up the frame condition from item 3, and the three rows now punctuate consistently.
> 
> ### Not changed
> Four lines sit at 101-102 columns. There is no enforced limit and `InitializePtcWindow.lean:47` is already 105, so I left them.
> 
> ### Verification after all seven
> `lake build` green, `just lint` clean, and `#print axioms` on all five theorems still reports only `propext`, `Classical.choice`, `Quot.sound`. No `sorry`, no new axiom, no mathlib.
> 
> Happy to drop any of these if you disagree, 2 and 3 are the two I would keep.

Thanks, Leo. I agree with the seven changes. 
2 and 3 are essential for avoiding namespace collisions and for stating the full frame condition. 
The other changes also make the proof structure and documentation clearer, and your explanations were very helpful for understanding the reasoning behind them.

Issue #14, opened 2026-07-06
SizzLeanTests: out-of-bounds panics during build


Mouzayan:
Running `lake build SizzLeanTests` reports out-of-bounds panics in:
- `SizzLeanTests/ElementSurface.lean`
- `SizzLeanTests/InfallibleIndex.lean`

The overall build still completes successfully:
```bash
Build completed successfully (166 jobs).
```

It is unclear whether these panics are expected behavior for these tests, or whether they indicate an issue in the tests or implementation.

**Reproduction**
`lake build SizzLeanTests`

**Observed output**
```bash
info: SizzLeanTests/ElementSurface.lean:72:0: PANIC at outOfBounds Init.GetElem:18:2: index out of bounds
info: SizzLeanTests/ElementSurface.lean:80:0: PANIC at outOfBounds Init.GetElem:18:2: index out of bounds
info: SizzLeanTests/InfallibleIndex.lean:97:0: Error: index out of bounds
info: SizzLeanTests/InfallibleIndex.lean:103:0: Error: index out of bounds
info: SizzLeanTests/InfallibleIndex.lean:114:0: Error: index out of bounds
info: SizzLeanTests/InfallibleIndex.lean:134:0: PANIC at outOfBounds Init.GetElem:18:2: index out of bounds
```

**Environment**
* OS: macOS / Apple Silicon
* Command: lake build SizzLeanTests

leolara:
Confirmed while reviewing #29, on `v4.29.1` / Linux. All six sites are deliberate out-of-bounds `!` operations, and every one of them asserts the documented fallback, so the panics are expected behavior.

The reads (`ElementSurface.lean`'s `xs[7]!` and `bs[7]!`, `InfallibleIndex.lean:134`'s `sszGet … xs[7]!`) go through `getElem!`, which routes an out-of-range index to `outOfBounds`. That panics on stderr and returns `default`. The gates assert exactly that value: `0` for `UInt64`, `false` for `Bool`.

The writes (`InfallibleIndex.lean:97`, `:103`, `:114`) are `sszUpdate … xs[7]! := 99`, which lands on `Array.set!`. It reports the miss and leaves the buffer unchanged, and the gates assert the resulting no-op on both the uncached and the cached spine.

[...]

PR #37, opened 2026-07-26
feat: implementing LeanHazmatXMSS package 

Sahilgill24:
## Summary

This PR implements the LeanHazmatXMSS package for test vector generation for XMSS signatures, it is the first step in implementing PQ XMSS signatures in etheorem, for the full detailed plan please have a look at the [XMSS-etheorem doc](https://hackmd.io/@Ayhm2-FHQhSLkoc1OEcr9g/HytVKthGGx)

## What changed

[WIP] 
Introduced a new Package LeanHazmatXMSS

## Test plan

  - [x] `just ethcl-test`

## LICENSE (LGPL version 3)


leolara:
Reviewed at `e5cfc40`. Thanks for this, the shape is right and the parts that were most likely to go wrong went right.

## What works well

The vendoring model is followed correctly. One C file is committed (`csrc/xmss_shim.c`, 224 lines), `vendor/` is gitignored, and the shim includes the vendored headers and calls `xmss_parse_oid` / `xmss_keypair` / `xmss_sign` / `xmss_sign_open` without reimplementing any crypto. The one function it does define, `randombytes`, is deliberate, since `randombytes.c` is excluded from the compiled source list.

The size accounting in the shim is worked out correctly and documented in the C header comment. The distinction between `XMSS_OID_LEN + params.pk_bytes` and bare `params.pk_bytes`, and the fact that `sig_bytes` carries no OID prefix, are both right. This is the classic place an XMSS binding breaks. Copying `sk` before `xmss_sign` mutates it correctly avoids writing to Lean-owned memory.

`Ffi.lean` follows the family's literate conventions closely: module docstring, per-declaration trust assumption, and `@[extern]` / `@&` / `opaque` explained on first use. The `extern_lib` archive shape, the gitignored `vendor/` tree, and `licenseFiles` all match the established pattern.

## What I ran

- `just hazmat-xmss-vendor`: clean. The pin `171ccbd` is currently upstream `master` HEAD.
- `lake build LeanHazmatXMSS`: green, 20 jobs.
- `lake build LeanHazmatXMSSTests`: green, all six `native_decide` examples fire.
- `just lint`: exit 0.
- Root `lake build` with `vendor/` removed: 0 jobs, so a missing vendor step does not break the umbrella build.

The PR is marked WIP, so below I have split what blocks merge from what should land before the WIP marker comes off.

## Blocking

### 1. The OID table in `LeanHazmatXMSS/Ffi.lean:15-17` has wrong signature sizes

I compiled a probe against the vendored `params.c` at the pinned commit:

```
oid=1  n=32  h=10  sig=2500      table says 2500   ok
oid=2  n=32  h=16  sig=2692      table says 4708   wrong
oid=3  n=32  h=20  sig=2820      table says 9124   wrong
```

Only the `_10_256` row is right. The API hands back concatenated buffers the caller splits by these numbers, so a caller trusting the table for `_16_256` slices garbage out of `sig ++ new_sk`.

### 2. `keygenSeeded` is declared `opaque` but is not a function

`csrc/xmss_shim.c:56-57` keeps `g_use_det_seed` and `g_det_counter` as unguarded process globals. Lean's kernel treats an `opaque` FFI result as pure. Two `native_decide` goals evaluated concurrently can interleave the counter and produce different keys for the same OID, which breaks the assumption the `opaque` declaration rests on.

Dropping the global is the clean fix. Expose `keygenFromSeed (seed : ByteArray)` and thread the seed down, rather than toggling a flag around `xmss_keypair`.

### 3. The archive exports a strong global `randombytes`

`csrc/xmss_shim.c:59`. `libleanhazmat_xmss.a` will collide with any other archive defining that symbol in the same executable, which a second PQ family or a future vendored library would hit. The `__libc_csu_*` stubs eight lines above are `weak` for exactly this reason. Same treatment here, or make it `static` behind a renamed wrapper.

## The tests do not test what the docs claim

### 4. There are no known-answer vectors

The module is called `Vectors.lean` and its docstring says "The KAT vectors are exactly what pins that trust", but no expected bytes are committed anywhere. All six examples round-trip through the same shim, so an XMSS implementation that is broken yet internally consistent passes every one of them.

`LeanHazmatBlsTests/Vectors.lean` is the model: published byte strings, then `sign skAnchor msgZero = sigAnchor`. RFC 8391 ships no official KATs, which makes this harder here than it was for BLS. Committing the deterministic-seed `pk` and one signature prefix as literals (or their SHA-256) still pins against regression, and would have caught finding 1.

### 5. The determinism example at `Vectors.lean:72-73` is vacuous

Both sides are the same syntactic expression, so `rfl` closes the goal regardless of what the FFI does at runtime. It cannot observe nondeterminism. Compare against a committed literal instead.

### 6. "Empty signature does not verify" at `Vectors.lean:66-69` exercises the C size guard

It never reaches signature validation. Flipping one byte of a valid `sig` is the negative test worth having.

## API design

**7.** Lean has no way to ask for `pk_bytes` or `sig_bytes`. The constants `68` and `2500` are retyped by hand in the test file. Something like `opaque paramSizes (oid : ByteArray) : Option (Nat × Nat × Nat)` removes the magic numbers and makes the API usable at any OID.

**8.** The `/dev/urandom` branch at `xmss_shim.c:66-70` is unreachable. No exported entry point ever enters xmss with `g_use_det_seed == 0`. Either expose a real keygen or delete it.

**9.** `smlen` is written by `xmss_sign` and never checked (`xmss_shim.c:156,160`). Assert `smlen == params.sig_bytes + msglen` before trusting the `sm` layout.

**10.** `keygenSeeded` does not sound as dangerous as it is. It returns a fixed, publicly derivable private key, and the docstring's caveat sits in the third paragraph.

## Naming

### 11. The package name breaks the family convention

Siblings are `LeanHazmatSha256`, `LeanHazmatBls`, `LeanHazmatKzg`: PascalCase acronyms. This one is `LeanHazmatXMSS`, while the PR's own Lean namespace is `LeanHazmat.Xmss` (`Ffi.lean:47`), so the package and its namespace disagree with each other.

`hazmat-docs/ARCHITECTURE.md` §3.2 makes the directory name the `git subtree split` prefix and the downstream mirror repo name. Renaming to `LeanHazmatXmss` now is far cheaper than renaming it after a mirror exists.

**12.** The test namespace is `LeanHazmatXMSSTests` (`Vectors.lean:30`); siblings use `LeanHazmat<Family>Tests.Vectors`.

## Documentation

### 13. `README.md` and `docs/ARCHITECTURE.md` are empty heading skeletons

`hazmat-docs/ARCHITECTURE.md` §13 makes the per-family doc a requirement: "Each family package also carries its own `docs/ARCHITECTURE.md` recording *its* trust boundary (library, version pin, validation vectors)." Every sibling has one at roughly 4 KB; this one is 154 bytes of headings.

It carries more weight here than it did for the other three. Upstream's own README calls the code "intended for cross-validation and experimenting" and warns specifically about deploying stateful signature schemes, and there are no published KATs to point at. BLS's doc could cite consensus-spec-tests v1.5.0. This one has to say what it pins trust to instead, and that belongs in writing before the package ships.

The README's Usage, API, and "Running and checking" sections are blank too.

### 14. `lakefile.toml:49-50` is factually wrong

The comment says "Self-contained: no OpenSSL dep, bundled SHA-256". `lakefile.lean:56` sets `moreLinkArgs := opensslLinkArgs`, and the lakefile's own header notes that `hash.c` calls OpenSSL's `SHA256()`.

### 15. The cross-family design of record was not updated

XMSS appears nowhere in `hazmat-docs/ARCHITECTURE.md` (539 lines), `hazmat-docs/PLAN.md` (517 lines), `docs/monorepo-arch.md`, the root `README.md`, or `CLAUDE.md`, which still says "Nine subpackages under `packages/`". Concretely:

`hazmat-docs/ARCHITECTURE.md`
- §5 Backends: a table of every family with backend, language, source, and rationale, deferred ones included. Needs an xmss-reference row, and the rationale column is where the reference-implementation caveat belongs.
- §12 layout: a tree listing every package directory, deferred ones marked as such. XMSS is absent.
- §14 Sequencing: two phases only, consensus and execution. A PQ signature family fits neither, so the table needs a third category.
- §7 and §8, the consensus and execution crypto surfaces, likewise have no slot for it.

`hazmat-docs/PLAN.md`
- Stages 0 through 10 are all allocated. XMSS needs its own stage.
- The Status snapshot table and the prose beneath it become wrong on merge: "Phase 1's families are complete. The consensus crypto surface ships as three à-la-carte packages" stops holding once a fourth family lands.

Per CLAUDE.md, ARCHITECTURE.md wins on substance, so these updates should land in the same PR as the package.

## Build

### 16. `lake-manifest.json` was reformatted by hand

The file went from single-line JSON objects to 2-space pretty-printed, turning one added package entry into a 150-line diff. CLAUDE.md forbids hand-editing it, and the next `lake update` anyone runs reverts the formatting. Regenerate it so the diff is just the new block.

**17.** `hazmat-xmss-vendor` does a full clone while its own doc-comment says "shallow clone at the pinned commit", and both sibling recipes use `--depth 1`. `git init && git fetch --depth 1 origin <sha> && git checkout FETCH_HEAD` works against GitHub. Since the pin is currently `master` HEAD, a `--depth 1` clone of the default branch plus a HEAD-equals-pin assertion also works.

**18.** `lakefile.lean:77-129` is nine near-identical five-line targets. Factoring the body into a helper that takes the source stem turns each into two lines. At 161 lines this is well past the 30-line guidance CLAUDE.md sets for procedural lakefiles; BLS does more in 98.

**19.** The pkg-config helpers were copied from `LeanHazmatSha256/lakefile.lean` with their docstrings stripped. The duplication itself is sanctioned by §3.3. The loss is that `opensslLinkArgs`'s `-L<libdir>` prepend exists to work around Lean's bundled `lld` not searching system paths, and without that comment it reads as redundant and gets deleted by the next reader.

**20.** CI was not updated. The `test`, `pyspec`, and `ethcl` jobs each run an explicit `just hazmat-bls-vendor hazmat-kzg-vendor`; xmss was not added. Nothing breaks, since root `lake build` is a no-op and `just test` self-vendors through `hazmat-xmss-test`. The effect is that CI now does a full clone in the middle of the test step instead of in the visible vendoring step.

**21.** Nits: trailing whitespace at `README.md:38`; blank line at EOF in `docs/ARCHITECTURE.md:24`; commit message typo `feat(hamzat-xmss)` should be `hazmat`; commit `80bd9e0` skips the conventional-commit prefix the rest of the history uses.

The four things I would fix first: the size table, real vectors, the two docs files, and the naming decision.

Sahilgill24:
hey @leolara 

I made all the changes that you had mentioned in your review along with some other changes that would have been needed,
The big one was keygen: it used to rely on a process-global counter to be deterministic, which is a real problem because two `native_decide` runs can race on it. I dropped that entirely and route keygen through xmss-reference's seeded entry point (`xmssmt_core_seed_keypair`), passing the seed down from Lean. Now the key is just a function of `(oid, seed)`, nothing hidden, and randombytes is only a weak stub so the archive still links. While in there I fixed the signature sizes in the OID table (they were wrong for the _16 and _20 sets), added paramSizes so callers can ask for the real `(pk, sig, sk) `sizes instead of copying constants around, and made sign check the length it gets back before slicing it.

 The tests were the other thing worth fixing. Before, they just signed and verified against output the shim itself produced, so a subtly broken implementation would still pass. Now they reproduce the same seed xmss-reference's own `test/vectors.c` uses and check the SHAKE128 digests that tool prints, for the public key and for a signature at index 512. If our shim drifts from upstream, this catches it. The SHAKE helper needed for that sits in its own Kat module that the package root doesn't export, so the actual API is still just the four signature calls. I also renamed the package LeanHazmatXMSS to LeanHazmatXmss to match the other hazmat packages and its own namespace.

Rest of the changes were basic housekeeping and nothing special, The one caveat I kept in writing is that this checks against xmss-reference's own vectors, since there aren't any official RFC 8391 ones to point at.

leolara:
One more item from reviewing the latest changes: `sign` on the last leaf returns a signature that does not verify.

This is a known upstream bug in xmss-reference. At `idx == 2^h - 1`, `xmssmt_core_sign` wipes the secret key and then signs with the wiped buffer. It still returns 0, so `lean_hazmat_xmss_sign` accepts the result and returns a bad `sig ++ new_sk`. Our pin `171ccbd` is `master` HEAD, so the bug is present. Upstream tracks it in https://github.com/XMSS/xmss-reference/issues/15, and the fix in https://github.com/XMSS/xmss-reference/pull/14 has been open and unmerged since 2021.

I reproduced it with upstream calls only, for XMSS-SHA2_10_256:

```
idx=1022 sign=0  verify=0     valid
idx=1023 sign=0  verify=-1    sign reports success, signature is invalid
idx=1024 sign=-2              refused
```

Please keep the vendored tree unmodified and:

1. In `lean_hazmat_xmss_sign`, return `mk_error()` when the index is `>= 2^h - 1`, before calling `xmss_sign`. Add a comment there that links issue #15 and PR #14.
2. Record the same in `docs/ARCHITECTURE.md` with both links, and state the usable capacity as `2^h - 1` signatures. The README says `2^h` today.
3. Add a test that sets the index to `0x000003FF` and expects `sign` to return the empty `ByteArray`.

leolara:
**`randombytes` stub** (`csrc/xmss_shim.c:67-73`)

The weak stub fills the buffer with zeros when `/dev/urandom` fails. Its comment also says a co-linked archive's `randombytes` "wins", which static linking does not guarantee. `xmss_shim.o` always links, so the weak definition is already in place. The linker then does not pull another archive's `randombytes.o`, and that library would call this stub instead.

Suggested fix: our API never reaches this path, so make the body `abort()`. Better still, compile the vendored sources with `-Drandombytes=lean_hazmat_xmss_ref_randombytes` and define only that name. Then the archive exports no global `randombytes` at all.

leolara:
**Unprefixed exported symbols**

`libleanhazmat_xmss.a` exports xmss-reference's globals under their bare names: `shake128`, `shake256`, `KeccakF1600_StatePermute`, `prf`, `hash_message`, `thash_f`, `wots_sign`, `ull_to_bytes`, and others. None collide today. A future Keccak family built on the common `fips202.c` would collide.

Suggestion: give every vendored global a package prefix, for example `lean_hazmat_xmss_ref_`. One way is a `csrc/xmss_prefix.h` with one `#define shake128 lean_hazmat_xmss_ref_shake128` line per symbol, force-included through `-include` in `xmssLibFlags` and `xmssShimFlags`. The vendored tree stays unmodified. `nm -g --defined-only` on the archive should then list only `lean_hazmat_xmss_*` names and the weak `__libc_csu_*` stubs.

leolara:
**Tests for the `_16` / `_20` sizes**

These sizes were the blocking bug last round, and no test pins them now. Please add:

```lean
example : paramSizes (.mk #[0, 0, 0, 2]) = some (68, 2692, 136) := by native_decide
example : paramSizes (.mk #[0, 0, 0, 3]) = some (68, 2820, 136) := by native_decide
```

Keygen at h = 16 and h = 20 is too slow for a build-time test, so these size checks are the only coverage those sets can get.

leolara:
**Docs and nits**

- **"One-time signatures" is wrong.** The phrase appears in `Ffi.lean`, the package README and `docs/ARCHITECTURE.md`, `lakefile.toml`, hazmat ARCHITECTURE §5, and PLAN Stage 11. XMSS is a stateful scheme that signs many messages. It is built on WOTS+ one-time signatures.
- **Regenerating the upstream digests.** Upstream does not commit them anywhere; `test/vectors.c` only prints them. Please add the command to `docs/ARCHITECTURE.md`: `cd vendor/xmss-reference && make test/vectors && ./test/vectors | grep '^XMSS 1 '`. I ran it, and it matches both committed digests.
- **Nits.** New em-dashes break the CLAUDE.md punctuation rule: the idioms list in `Ffi.lean`, two places in `lakefile.lean`, the XMSS Justfile comments, and the CI comment. The `hazmat-xmss-test` description still says "round-trip KAT". The subjects `feat(hamzat-xmss)` and `LeanHazmatXMSS: package initialized…` are still in the history; a squash-merge removes them.

PR #42, opened 2026-08-03
feat(ethcllib): isValidMerkleBranch completeness over openable paths

IvanAnishchuk:
## Summary

`isValidMerkleBranch` accepts the honest opening of any tree it can open. Open a
tree at `index`, hand the check the leaf and the sibling path, and the check
returns `true`. No behaviour change.

Builds on #71 and #72.

## What changed

- `IsOpenable`, `leafAt`, `siblingPath`, `honestLeaf` and `honestBranch`, with
  their size lemmas, in a new `EthCLLib/Proofs/MerkleOpening.lean`.
- `isOpenable_of_isPerfect`.
- `isValidMerkleBranch_complete`, with corollaries for a perfect tree and for
  `Sha256Spec`.
- `Tests/MerkleWitness.lean`: two rejection witnesses for the length guard, one
  branch too long and one too short.

## Test plan

  - [x] `just ethcl-test`
      `just ethcl-pyspec` is green.
- [ ] If this touches a bench-measured hot path: included
      before/after `lake exe ssz_bench` TSV (or relevant rows) in
      the PR description.
- [x] New SSZ shapes / cache cases come with a `native_decide` or

## Risk / trust footprint

- Adds no axiom directly. `isValidMerkleBranch_complete` needs no
  `CombineWidth32` instance, because `IsOpenable` records the sibling widths
  itself.
- The perfect-tree corollary takes a `CombineWidth32` instance. At `Sha256` that
  resolves to the instance proved from `sha256Combine_eq_spec`, so an FFI
  instantiation inherits that axiom. The `Sha256Spec` corollary resolves with no
  axiom.
- Adds and removes no `sorry`, `partial def`, `native_decide`, `unsafe` block
  or `@[extern]` declaration.

## Notes for reviewers

The openers sit in EthCLLib while shape and width stay in SizzLean. The routing
convention belongs to `is_valid_merkle_branch`.

The injection theorem takes the name `isOpenable_of_isPerfect` instead of
`IsPerfect.isOpenable`, which keeps an EthCLLib-typed result out of SizzLean's
namespace.

On a shape mismatch, `leafAt` and `siblingPath` return a zero chunk and report
nothing. A caller therefore depends on holding the `IsOpenable` witness.

## LICENSE (LGPL version 3)

### DIFF
--- a/packages/EthCLLib/EthCLLib/Proofs.lean
+++ b/packages/EthCLLib/EthCLLib/Proofs.lean
@@ -1,3 +1,4 @@
+import EthCLLib.Proofs.MerkleOpening
 import EthCLLib.Proofs.MerkleBranch
 
 /-!
--- a/packages/EthCLLib/EthCLLib/Proofs/MerkleBranch.lean
+++ b/packages/EthCLLib/EthCLLib/Proofs/MerkleBranch.lean
@@ -1,4 +1,8 @@
 import EthCLLib.Spec.SigningRoot
+import EthCLLib.Proofs.MerkleOpening
+import SizzLean.Proofs.Perfect
+import SizzLean.Hasher.CombineWidth
+import SizzLean.Proofs.Util
 
 /-!
 # `EthCLLib.Proofs.MerkleBranch`: what `isValidMerkleBranch` accepts
@@ -9,9 +13,24 @@ import EthCLLib.Spec.SigningRoot
 Past the length guard, the check reduces to a plain fold of `branch` over `leaf`,
 compared to `root`.
 
+Open a tree at `index` with `honestLeaf` / `honestBranch`, and the check returns
+`true`. `IsOpenable` (`EthCLLib/Proofs/MerkleOpening.lean`) is the path-shaped
+hypothesis, and every `IsPerfect` tree satisfies it at every index
+(`isOpenable_of_isPerfect`).
+
+The theorems quantify over `Node.merkleRoot`, the tree's own root. The spec's
+callers pass a root from elsewhere: `process_deposit` reads
+`state.eth1_data.deposit_root`, and hands the check a wire-deserialised `Array`
+rather than a `Node`.
+
 Bit ordering: `branch` runs bottom-to-top and entry `i` is the level-`i` sibling.
 The routing convention lives in `EthCLLib.Spec.MerklePath`. The shipped check and
 this fold both call `routeRight`, so the two agree by definition.
+
+Under an abstract `[HasherTag]`, `isValidMerkleBranch_complete` and the two guard
+bridges apply as they stand. The results carrying `[CombineWidth32 HasherTag.H]`
+need `pureHasherTag` or `fastHasherTag`, since `HasherTag.H` is otherwise a stuck
+projection matching no instance.
 -/
 
 set_option autoImplicit false
@@ -19,6 +38,9 @@ set_option autoImplicit false
 namespace EthCLLib.Proofs
 
 open SizzLean
+open SizzLean.Hasher
+open SizzLean.Cache.MerkleTree
+open SizzLean.Proofs
 open EthCLLib.Spec
 
 /-! ### The proof-side fold
@@ -47,6 +69,19 @@ private theorem branchFold_succ [HasherTag] (leaf : ByteArray)
   unfold branchFold
   rw [List.range_succ, List.foldl_append]
   rfl
+
+/-- **Reads at or above `depth` cannot matter.** Two branches agreeing below
+`depth` drive the same walk. The induction step can therefore ignore the sibling
+`siblingPath` pushes on at the top. -/
+private theorem branchFold_congr [HasherTag] (leaf : ByteArray)
+    (b b' : Array (Vector UInt8 32)) (depth index : Nat)
+    (hagree : ∀ i, i < depth → b[i]! = b'[i]!) :
+    branchFold leaf b depth index = branchFold leaf b' depth index := by
+  induction depth with
+  | zero => rfl
+  | succ d ih =>
+      rw [branchFold_succ, branchFold_succ, ih (fun i hi => hagree i (by omega)),
+          hagree d (by omega)]
 
 /-- The checked fold itself, in range: every `branch[i]` read hits, so the
 `IndexError` arm never fires. The statement sits on the inner `foldlM`, since the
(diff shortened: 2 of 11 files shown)


leolara:
@IvanAnishchuk this is in draft, do you want me to review it already?

IvanAnishchuk:
Not yet, work in progress, wanted to rework a couple things I'm completely unsure of first

leolara:
@IvanAnishchuk Great job proving!!

A first quick look, what surprised me is why if isValidMerkleBranch is not related to SSZ, why we needed to add things there. I see we add merkleFold to the SSZ library, but this function is not related to SSZ objects or structures. All the merkleFold machinery should be in the EthCLLib, I think.

leolara:
## Proposal: split this branch into a stack

Three fairly separable pieces sit in here, and the proof half is what makes the
review expensive. I'd like to land them as three PRs instead, in this order.

### 1. `is_valid_merkle_branch` and its length guard

- `EthCLLib/Spec/SigningRoot.lean`: `merkleBranchRoot` and the dependent-`if` guard.
- `routeRight` and `merkleFold` with their lemmas, authored in EthCLLib rather than
  SizzLean. More on that under "Where the branch convention lives".
- The three flat lemmas at `EthCLLib/Proofs/MerkleBranch.lean:147-196`
  (`isValidMerkleBranch_eq_beq`, `branch_size_of_isValidMerkleBranch`,
  `isValidMerkleBranch_iff`). Their whole dependency closure is the check and
  `merkleBranchRoot`, so they bring no tree vocabulary with them.
- The `EthCLLibProofs` `lean_lib`, its `just ethcl-test` line and the CI job.
- The `DISCREPANCIES.md` and `IMPLEMENTATION_NOTES.md` entries on
  `is_valid_normalized_merkle_branch`.

Roughly 350 lines, and the only behavior change on the branch. A short branch used
to fold `Inhabited` defaults in through `branch[i]!` where upstream raises
`IndexError`, and now it rejects. The conformance sweep makes most sense here, since
this is the only piece that can move a vector.

### 2. The theorems connecting the check to the SSZ Merkle tree

Staying in SizzLean, since they are facts about the tree the library already owns:

- `Node.rootOf` going structural, the six equation lemmas, and the `@[irreducible]`
  seals on `rootOf` and `merkleRoot`.
- `CombineWidth32` and its two instances.
- `IsPerfect` and its shape facts: `merkleRoot_size`, `merkleRoot_ofSubtrees_size`,
  `rootOf_eq_merkleRoot`, `zeroLeaf_isPerfect_*`, `ofLeaves_isPerfect_le`.
- `Proofs/Util.lean` and `natToChunk_size`.

Moving to EthCLLib, since each one addresses a tree by the branch convention and so
depends on `routeRight`:

- `IsOpenable`, `leafAt`, `siblingPath`, and their size lemmas.
- The three witness producers: `IsPerfect.isOpenable`, `mixInLength_isOpenable`,
  `ofSubtrees_isOpenable`.
- `Proofs/MerkleOpening.lean` and the completeness half of `Proofs/MerkleBranch.lean`.

Plus the `ofSubtrees` truncation `#guard`. Roughly 1700 lines.
`Tests/MerkleWitness.lean` can split off as a further PR on top if this still reads
as too much in one go.

The line falls where the vocabulary does. SizzLean answers "this tree is perfect and
this root is 32 bytes", EthCLLib answers "this is the node a branch addresses".

### 3. The merkleizer changes

`ZERO_HASH_TABLE_DEPTH`, the `zero32` dedup, threading the zero tower through
`merkleize`, and the switch from clamping to panicking on an out-of-range depth. No
theorem in (2) mentions `zeroHashOf` or `ZERO_HASHES_SPEC`, so this stands on its
own and reviews as a correctness plus performance change.

### Ordering

(2) needs `ZERO_HASH_TABLE_DEPTH`, which the `IsPerfect` half states twelve times.
So either (3) goes first, or the constant and the `zero32` dedup ride along in (2)
and (3) keeps the threading and the panic.

### Where the branch convention lives

I think `merkleFold` and `routeRight` belong in EthCLLib. Every caller is there, no
SizzLean code calls either one, and the module docstring points at
`consensus-specs/specs/phase0/beacon-chain.md:782-798` while explicitly ruling out
the SSZ document's generalized-index verifier. A beacon-chain function sits
awkwardly in the namespace that holds `Serialize`, `Deserialize`, and
`HashTreeRoot`. The signatures agree: `ByteArray`, `Nat`, and the `Hasher` seam,
with no SSZ type in sight.

That also decides how (2) splits, since SizzLean cannot import EthCLLib and
`leafAt` / `siblingPath` / `IsOpenable` all route on `routeRight`. Worth settling in
(1), so we don't pay for a second round of import churn later.

### Two smaller things I noticed

Both are easy to handle while the code is moving anyway.

**`get!_eq_getElem` is declared three times under one name.** The new
`SizzLean/Proofs/Util.lean:28` copy shares its fully qualified name with
`Proofs/BitPack.lean:84` and `Proofs/UIntWide.lean:90`. Imports survive it because
the statements are identical and Lean's `subsumesInfo` check
(`Lean/Environment.lean:2229`) lets one stand in for another; the environment then
attributes the constant to whichever module loaded first, which is BitPack. So
`MerkleOpening.lean:71` is citing BitPack's copy today. Change any one of the three
statements and the import fails with *environment already contains*. The collision
predates this branch, and folding all three into `Proofs/Util.lean` would close it.

**Ten declarations currently have no consumer.** Nine of them are reachable only
through `IsOpenable.trans`, which nothing calls anywhere in the repo:
`index_congr`, `routeRight_combine_low` and `_high`, `routeRight_eq_testBit`,
`routeRight_eq_div_mod`, `nodeAt`, `nodeAt_zero`, `leafAt_eq_rootOf_nodeAt`.
`ofLeaves_isPerfect_le` is separate, a fourth `IsPerfect` producer that no corollary
reaches. `le_two_pow_chunkDepth` has exactly one caller,
`Tests/MerkleWitness.lean:337`. My inclination is to hold the unused ones back and
bring them in with whatever first consumes them, though `routeRight_eq_div_mod` is
worth keeping regardless, since it pins the convention against upstream's literal
spelling.

Both lists came out of the compiled environment rather than the source: for each
constant, the union of its type's and value's `getUsedConstants`, contracted through
internal declarations and filtered to the modules this branch touches.

IvanAnishchuk:
Yeah, splitting it into stack was what I had in mind. Did it partially + cleaned up this branch. Let me see if I can split it any further and address other parts of the review...

IvanAnishchuk:
Okay, I think I'm done splitting and cleaning up prose language. Not sure if my STE-ASD linter was very effective, let me know if prose still reads too weird, I'll go over it with a full semantic rewrite if needed. Upper two branches of the stack are still big, perhaps I should do a similar splitting with them too.

I'll self-review it all one last time and start marking ready to review.

leolara:
Please, can you explain which PRs are which part of the split and in which order I should review them?

IvanAnishchuk:
@leolara sorry, I should've not created the PRs from my fork (stacking feature doesn't support that), this way PRs up in the stack accumulate commits and a hard to review - I can close and recreate but seemed like too much hussle

1. #79 - is a small preliminary fix to make axiom audit work

2. #71 - `compute_merkle_branch_root` and the length guard
3. #72 - perfect combine-trees
4. #42 - `isValidMerkleBranch` completeness

(these three are roughly the content of the original completeness branch split)

6. #66 - generalized indices
7. #67 - `hash_tree_root` agreement

IvanAnishchuk:
rebased the stack

IvanAnishchuk:
Closing this. #100 landed the pure-path merkleization, gindex, and branch results, and it set the cached-tree proofs aside as out of scope. The pure-path parts this stack still adds are rebased onto `main` as #111 and #112. The rest was either superseded by #100 or belongs to the cached path.

Issue #68, opened 2026-08-18
Prove the three central theorems for vector/list over a variable-size element type (vectorVar / listVar)


irajgill:
## Motivation

The codec already serializes and deserializes `.vector t n` / `.list t cap` when `t` is variable-size (`list`, `bitlist`, mixed-field `container`, nested variable collections):

- `Spec/Serialize.lean`'s `serializeVarElemsAux` writes the offset-table wire format: `n` uint32 placeholders, then concatenated bodies, with the running offset seeded at `n * BYTES_PER_LENGTH_OFFSET`.
- `Spec/Deserialize.lean`'s non-`isFixedSize` branches call `extractCollOffsets` then `deserializeVarElems`. Lists recover `count` from `firstOff / 4`; the empty list is the empty buffer. Vectors still reject `n = 0`.

What is missing is the proof surface. `Supported` / `BasicSupported` only have `vectorFixed` / `listFixed` (`t.isFixedSize = true`). `Supported.vectorFixed`'s docstring still says the variable-size offset-table read is not implemented; that is stale. README / `docs/PLAN.md` Stage 18 name this as the remaining composite gap after mixed-field containers (#27 / #28 / #29).

This is the August cell of the EPF joint proposal: close the last codec-implemented composite arm so the three central theorems cover every type the encoder and decoder actually handle (modulo the schema-level `MAX_LENGTH` guard, #61).

## Scope

Add `vectorVar` / `listVar` constructors to `SSZType.Supported`, `SSZType.SupportedBounded`, and `SSZType.BasicSupported` (alongside the existing `*Fixed` constructors, not replacing them), and prove `decode_encode`, `serialize_injective`, `encode_size_le_max` for both. `serialize_injective` inherits from `decode_encode`.

### Preconditions on `BasicSupported.vectorVar t n`

- `0 < n` (decoder rejects empty vectors, same as `vectorFixed`)
- `BasicSupported t`
- `t.isFixedSize = false` (routes encode/decode onto `serializeVarElemsAux` / `deserializeVarElems`)
- `maxByteLength (.vector t n) < MAX_LENGTH` after the bound fix below (uint32-offset overflow guard; running offsets start at `n * 4` and grow by body sizes)

### Preconditions on `BasicSupported.listVar t cap`

- `BasicSupported t`
- `t.isFixedSize = false`
- `maxByteLength (.list t cap) < MAX_LENGTH` after the bound fix (same overflow guard; empty list is in-bounds)

[...]

PR #70, opened 2026-08-18
feat(sizzlean): groundwork for variable-element vector/list proofs

irajgill:
## Summary

Second of three PRs for #68. This PR adds proof groundwork in `Proofs/CollectionVar.lean`. Predicate and dispatcher changes remain in #75, so this slice builds on its own.

#69 is already on `main`. Homogeneous collections have one element type `t`. The proofs induct on the element list and reuse the existing uint32 codec bridge from `ContainerVar`.

- **Offset-list plumbing** (`collOffsetsOf`): records the running offsets written by `serializeVarElemsAux`.
- **Encoder accounting** (`size_serializeVarElemsAux_offs`): proves that the offset table has size `xs.length * 4`. At the encoder seed, the first encoded offset is `n * 4`. The vector decoder takes `count` from `n`. Roundtrip uses the encoder's canonical first offset. Decoder canonicity remains outside #68.
- **First-offset read** (`readUInt32LE_serializeVarElemsAux_cons`): proves that the list decoder reads the seeded `varOff`. Together with `toNat_toUInt32_of_lt`, this recovers `count = firstOff / 4`.
- **Size walker** (`size_serializeVarElemsAux_le_max` and its collection specializations): bounds the encoded size using the formula from #69. The later roundtrip proofs use this bound for uint32 offsets.
- **Offset-extraction inverse** (`extractCollOffsets_serializeVarElemsAux`): proves that extraction from encoder output recovers the running offsets.

The PR also changes `extractCollOffsets` from `private` to `protected`, matching `extractFieldOffsets`.

## Verification

- `lake build SizzLean SizzLeanTests`
- Axiom audit on every new theorem
- Accounting and size lemmas use only their standard kernel dependencies
- Placeholder-reading lemmas reuse one existing `bv_decide` axiom from the uint32 bridge

## Test plan

- [x] SizzLean and SizzLeanTests build
- [x] New theorem axioms checked
- [x] CI passes
- [x] #69 is on `main`

## LICENSE (LGPL version 3)

### DIFF
--- a/packages/SizzLean/SizzLean/Proofs/CollectionVar.lean
+++ b/packages/SizzLean/SizzLean/Proofs/CollectionVar.lean
@@ -20,12 +20,11 @@ a follow-up, once this file's lemmas are available to build on and
 once `containerVar` is on `main` (the `BasicSupported` matchers grow
 two constructors; that arm must already be exhaustive).
 
-Homogeneous collections, so this is closer to `Proofs/FixedElems.lean`
-than to `Proofs/ContainerVar.lean`: one element type `t`, induction
-on the element list. No new heterogeneous mutual predicate. The
-uint32 codec bridge (`readUInt32LE_uint32LE_append`,
-`readUInt32LE_append_shift`, `toNat_toUInt32_of_lt`) is reused from
-`ContainerVar` rather than re-proved.
+Homogeneous collections have one element type `t`, so the proof
+inducts on the element list as `Proofs/FixedElems.lean` does. The
+existing uint32 codec bridge from `ContainerVar`
+(`readUInt32LE_uint32LE_append`, `readUInt32LE_append_shift`,
+`toNat_toUInt32_of_lt`) supplies the byte-level inverse.
 
 ## Wire format (what the lemmas below characterize)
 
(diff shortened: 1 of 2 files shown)


leolara:
Reviewed the rebased head (`09a0fe2`). The module builds clean, has no
`sorry`, and the axiom footprint matches the description: the accounting and
size lemmas use only `propext` + `Quot.sound`, and the two offset-reading
lemmas add `Classical.choice` plus the one `bv_decide` certificate already
inherited from `ContainerVar`.

Every lemma statement checks out against the code. `collOffsetsOf` mirrors
`serializeVarElemsAux`, the `varOff` seed of `xs.length * BYTES_PER_LENGTH_OFFSET`
matches both encoder call sites (`Spec/Serialize.lean:320`, `:331`), and the
two `maxByteLength` specializations match the `.vector` / `.list` non-fixed
arms. The `private` to `protected` change on `extractCollOffsets` is complete
and follows the `extractFieldOffsets` convention.

One thing to raise.

**A decoder gap that blocks the proofs this PR is groundwork for. Filed as #81.**

Neither collection arm anchors the first offset to the size of the offset
table, so both accept `off₀ = 0`, which points back into the table. The
container arm does check this (`Deserialize.lean:433`). The consequence is
that `serialize (deserialize b) = b` fails on buffers the decoder accepts.

This matters here specifically: the `vectorVar` and `listVar` arms of
`decode_encode` and `serialize_injective`, the theorems #68 tracks and these
lemmas feed, cannot be stated truthfully until the decoder rejects those
buffers. The property they would assert is false today. #81 has the failing
cases as runnable `native_decide` examples.

Worth folding the fix into this PR rather than filing it forward, since the
proof work is blocked on it either way.

leolara:
I edited the comment above to drop a second point about
`set_option maxHeartbeats 4000000`. I thought the raise was unwarranted. I
was wrong. Every module in `SizzLean/Proofs/` raises the limit, and 4000000
matches `ContainerVar`, which this proof is modelled on.

PR #67, opened 2026-08-18
feat(sizzlean): hash_tree_root agreement, and the gindex results over spec roots

IvanAnishchuk:
## Summary

The cache tree's root agrees with `SSZType.hashTreeRoot`, and the
generalized-index results now state their conclusion over that root. No
behaviour change.

Stacked on #66.

## What changed

- `SizzLean/Proofs/Merkleize.lean` (new): `merkleize_eq_ofLeaves`,
  `rootOf_ofSubtrees_eq_ofLeaves`.
- `SizzLean/Proofs/ShapeAgreement.lean` (new):
  `merkleRoot_ofShape_eq_hashTreeRoot`.
- `SizzLean/Spec/HashTreeRoot.lean`: unseals `merkleize`, `zeroHashAtDepth`,
  `combineLayerAt` and `promoteThroughZeros`, adds `merkleizeAt` with
  `merkleize_eq_merkleizeAt`, and the equation lemmas for all of them.
- `SizzLean/Spec/Interp.lean`: `SSZType.fieldAt`.
- `SizzLean/Cache/MerkleTree/Zero.lean`: unseals `zero32`, adds
  `zeroHashAt_zero`.
- `EthCLLib/Proofs/MerkleShape.lean` (new): `nodeAt_ofSubtrees`,
  `nodeAt_ofShape_container`.
- `EthCLLib/Proofs/GeneralizedIndexBranch.lean`:
  `isValidMerkleBranch_complete_containerField_specRoot`. The two-step witness
  takes `hfield` and `hval` in place of `hsub`.

## Test plan

- [x] `lake build` (full monorepo, on the pinned toolchain) is green.
- [x] Per-package test suites pass:
  - [x] `lake build LeanSha256Tests`
  - [x] `lake build SizzLeanTests`
  - [x] `just ethcl-test`
- [x] If this touches spec types or cache behaviour:
      `just ethcl-pyspec` is green.
- [ ] If this touches a bench-measured hot path: included
      before/after `lake exe ssz_bench` TSV (or relevant rows) in
      the PR description.
- [x] New SSZ shapes / cache cases come with a `native_decide` or
      property test that exercises them.

## Risk / trust footprint

- Unseals four definitions in `Spec/HashTreeRoot.lean` and `Cache.zero32`, so
  tactics may now unfold them.
- Adds and removes no axiom, `sorry`, `partial def`, `native_decide`, `unsafe`
  block or `@[extern]` declaration.

## Notes for reviewers

[...]

PR #66, opened 2026-08-18
feat(ethcllib): tie generalized indices to the tree they address

IvanAnishchuk:
## Summary

`get_generalized_index` over `SSZType`, and `isValidMerkleBranch` completeness
at the indices it returns for one- and two-step container paths. No behaviour
change.

Stacked on #42.

## What changed

- `SizzLean/Spec/GeneralizedIndex.lean` (new): `getGeneralizedIndex` over a
  `List PathStep` for all seven `SSZType` arms, with `chunkCount`,
  `itemPosition` and `elemType`. Plus
  `getGeneralizedIndex_container_field`, `floorLog2_two_pow_add` and
  `getSubtreeIndex_two_pow_add`.
- `SizzLean/Proofs/ShapeWidth.lean` (new): `merkleRoot_ofShape_size`.
- `SizzLean/Proofs/Perfect.lean`: `zeroLeaf_isPerfect_of_combine`,
  `zeroLeaf_isPerfect_sha256`, `merkleRoot_ofSubtrees_size`,
  `cached_root_honest`.
- `EthCLLib/Proofs/MerkleOpening.lean`: `nodeAt`, `nodeAt_zero`,
  `routeRight_combine_low` and `_high`, `IsOpenable.index_congr`,
  `IsOpenable.trans`, `ofSubtrees_isOpenable`.
- `EthCLLib/Spec/MerklePath.lean`: `routeRight_eq_testBit`.
- `EthCLLib/Proofs/GeneralizedIndexBranch.lean` (new):
  `isValidMerkleBranch_complete_gindex`, `_containerField`, `_containerField₂`,
  `ofShape_container_isOpenable`.

## Test plan

- [x] `lake build` (full monorepo, on the pinned toolchain) is green.
- [x] Per-package test suites pass:
  - [x] `lake build LeanSha256Tests`
  - [x] `lake build SizzLeanTests`
  - [x] `just ethcl-test`
- [x] If this touches spec types or cache behaviour:
      `just ethcl-pyspec` is green.
- [ ] If this touches a bench-measured hot path: included
      before/after `lake exe ssz_bench` TSV (or relevant rows) in
      the PR description.
- [x] New SSZ shapes / cache cases come with a `native_decide` or
      property test that exercises them.

## Risk / trust footprint

- `zeroLeaf_isPerfect_sha256` inherits `sha256Combine_eq_spec` through
  `zeroHashAt_size`.
- Adds and removes no `sorry`, `partial def`, `native_decide`, `unsafe` block or
[...]

PR #71, opened 2026-08-19
feat(ethcllib): compute_merkle_branch_root and the length guard

IvanAnishchuk:
## Summary

`isValidMerkleBranch` rejects a `branch` whose size differs from `depth`. The
walk over `branch` moves into `computeMerkleBranchRoot`, which reads each
sibling with a bounds check. That guard is the only behaviour change here.

## What changed

- `computeMerkleBranchRoot`, the pyspec's `compute_merkle_branch_root`
  (`beacon-chain.md:782-798`). Out-of-range sibling reads raise `IndexError`.
- `isValidMerkleBranch` guards on `branch.size = depth`.
- `routeRight` and `routeRight_eq_div_mod`, in a new
  `EthCLLib/Spec/MerklePath.lean`.
- `isValidMerkleBranch_eq_beq`, `isValidMerkleBranch_iff` and
  `size_map_bytesToRoot`.
- `get!_eq_getElem` moves to `SizzLean/Proofs/Util.lean`. The `BitPack.lean` and
  `UIntWide.lean` copies go.

## Test plan

  - [x] `just ethcl-test`
      `just ethcl-pyspec` is green.
- [ ] If this touches a bench-measured hot path: included
      before/after `lake exe ssz_bench` TSV (or relevant rows) in
      the PR description.
- [x] New SSZ shapes / cache cases come with a `native_decide` or

## Risk / trust footprint

None. This pull request adds and removes no axiom, `sorry`, `partial def`,
`native_decide`, `unsafe` block or `@[extern]` declaration.

## Notes for reviewers

`routeRight` sits in EthCLLib and not in SizzLean. `is_valid_merkle_branch` is a
consensus-spec function, and the SSZ document defines no such convention.

## LICENSE (LGPL version 3)

### DIFF
--- a/packages/EthCLLib/EthCLLib.lean
+++ b/packages/EthCLLib/EthCLLib.lean
@@ -1,6 +1,7 @@
 import EthCLLib.Internal.Capture
 import EthCLLib.Internal.ProofLedger
 import EthCLLib.Spec
+import EthCLLib.Proofs
 import EthCLLib.PySpecTests.Interface
 import EthCLLib.PySpecTests.Driver
 
--- a/packages/EthCLLib/EthCLLib.lean
+++ b/packages/EthCLLib/EthCLLib.lean
@@ -17,7 +17,8 @@ the generic `PySpecTests` driver. It names no fork; a separate package
 
 The author-facing surface is gathered under `EthCLLib.Spec`, so a spec file
 opens exactly one namespace (`open EthCLLib.Spec`). Internals live under
-`EthCLLib.Internal`; the generic pyspec driver under `EthCLLib.PySpecTests`.
+`EthCLLib.Internal`; the generic pyspec driver under `EthCLLib.PySpecTests`;
+the theorems about the `Spec` functions under `EthCLLib.Proofs`.
 
 One internal module is author-facing all the same: `Internal.ProofLedger` defines
 the `characterizes` attribute a proof carries to claim it states a spec
--- a/packages/EthCLLib/EthCLLib/Proofs.lean
+++ b/packages/EthCLLib/EthCLLib/Proofs.lean
@@ -1,0 +1,5 @@
+import EthCLLib.Proofs.MerkleBranch
+
+/-!
+# `EthCLLib.Proofs`: framework proof modules
+-/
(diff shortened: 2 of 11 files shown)


leolara:
Rebased on `main`, with seven commits on top. The rebase dropped the `fix(proof-coverage)` commit, already on `main`.

Changes from the original description:

- **`private` theorems.** `isValidMerkleBranch_eq_beq` and `isValidMerkleBranch_iff` state their right-hand side with `branchFold`, which is `private`. Outside the module that helper has no name, so no goal could match either theorem. Both are `private` now.
- **`Proofs/Util.lean` header.** It said three modules declared `get!_eq_getElem` before. Two did, `Proofs/BitPack.lean` and `Proofs/UIntWide.lean`. It also named the `Cache/MerkleTree` modules as importers, and none import it. Both claims are corrected.
- **Prose.** Contrastive negation in three places, one citation format (`phase0/beacon-chain.md:<span>`, replacing a bare span that named no file), and the figures of speech in `MerklePath.lean`. `EthCLLib/Proofs.lean` gained a real module docstring.
- **`PROOF_LEDGER.md`.** The `processDeposit` dependency pointed at SizzLean PLAN Phase 5 for the whole branch check. It names `EthCLLib/Proofs/MerkleBranch.lean` now, and keeps Phase 5 for the merkleization agreement.
- **Tests.** Three cases in `Tests/FrameworkUtils.lean` for the length guard and the `IndexError` arm. `isValidMerkleBranch r0 #[r0] 0 0 r0` is the one that earns its keep: the old defaulting read accepted it.
- **One dead `open SizzLean.Hasher`** removed from `MerkleBranch.lean`.

Green on the rebased branch: `lake build`, `just ethcl-test`, `just lint`, `just check-citations`, `just proof-coverage-check` (41 claims), `just ethcl-pyspec` (152 passed), and every deposit vector at fulu/minimal (683 passed).

Two things left for you: the description above still describes the pre-rebase state, and PRs #72, #42, #66 and #67 carry the old text of `MerkleBranch.lean`, `MerklePath.lean` and `Util.lean`, so they need a rebase.

PR #82, opened 2026-08-23
feat(heze-focil-verdict-recording-proof): prove recordPayloadInclusionListSatisfaction successful-run write

Mouzayan:
## Summary

Adds `EthCLSpecs/Proofs/Heze/RecordPayloadInclusionListSatisfaction.lean`,
proving the successful-run write of Heze’s FOCIL satisfaction-result recorder
at `ForkChoiceStoreRun (Store map)`:

- Given a nonzero `state.slot`, if timely
  `getInclusionListTransactions` collection runs from `runnerStore` and returns
  `ilTxs` with runner state `postRunnerStore`,
  `recordPayloadInclusionListSatisfaction` returns the explicit input store
  with `payloadInclusionListSatisfaction` updated by `FcMap.insert` of
  `isInclusionListSatisfied payload ilTxs` at `root`.
- The recorder threads through exactly the `postRunnerStore` produced by
  transaction collection; it does not independently modify the runner state.
- The equation covers both Boolean satisfaction results.

## Supporting Changes

- Re-exports the module from `EthCLSpecs/Proofs/Heze.lean`.
- Records the successful-run frame theorem in `PROOF_LEDGER.md`.
  `recordPayloadInclusionListSatisfaction` remains **in progress** because its
  failure branches are not characterized.
- Keeps the same-root lookup and payload/result pairing obligation on the
  `onExecutionPayloadEnvelope` row. `shouldExtendPayload` also remains in
  progress.
- Updates the Proofs section of `IMPLEMENTATION_NOTES.md`.
- Regenerates the proof-coverage baseline and README blocks with
  `just proof-coverage-update`.

## Notes for Reviewers

- This is an exact successful-run write/frame theorem, not a complete
  characterization of every execution. It excludes the slot-zero
  `checkedSub` failure and transaction-collection failures, including the
  empty-committee and missing-timeliness-key paths.
- The explicit input `store` and runner state are distinct. The function
  returns an updated `Store` as its monadic value while threading through the
  runner state produced by transaction collection.
  `onExecutionPayloadEnvelope` is responsible for installing the returned
  store.
- The transaction-collection hypothesis explicitly pins
  `(onlyTimely := true)`, matching the implementation default.
- The theorem proves the structural `FcMap.insert` update. It does not claim
  that subsequently looking up `root` returns the computed satisfaction
  result, because generic `[FcMap map]` provides no insert/lookup law. That is
  a limitation of the map abstraction, not of the frame proof; the associated
  lookup and same-root pairing obligation remains on
  `onExecutionPayloadEnvelope`.
- The theorem records the result returned by
  `isInclusionListSatisfied`; it does not prove that the execution engine’s
  result is semantically correct.
- The proof uses targeted `simp` followed by `rfl`.
- The elaborated theorem requires `[Heze.Preset]`, `[HasherTag]`,
  `[FcMap map]`, and
  `[ExecutionEngine ExecutionPayload Transaction ExecutionRequests]`.
- `#print axioms` reports exactly `propext`, `Classical.choice`, and
  `Quot.sound`. No Mathlib, `sorry`, `native_decide`, or `bv_decide` is used.

## Test Plan

- [x] `lake build`
- [x] `lake build EthCLSpecs`
- [x] `just ethcl-test`
- [x] `just lint`
- [x] `just test`
- [x] `git diff --check`
- [x] `just check-citations`
- [x] `just proof-coverage-update`
- [x] `just proof-coverage-check`
- [x] `#print axioms` on
  `recordPayloadInclusionListSatisfaction_run_eq`

## Risk / trust footprint

Uses the existing kernel axioms `propext`, `Classical.choice`, and
`Quot.sound`. Adds no `sorry`, `partial def`, `native_decide`, `bv_decide`,
`unsafe`, or `@[extern]`.

## LICENSE (LGPL version 3)

- [x] I have read and accepted LICENSE (LGPL version 3), NOTICE, and CLA in the
  root of this project.


leolara:
@Mouzayan I think we should try to make the text in comments and markdown more understandable

leolara (approved):

PR #83, opened 2026-09-01
feat(heze-focil): characterize inclusion-list collection and verdict recording

Mouzayan:
## Summary

Completes the `.run` characterization of Heze’s inclusion-list collection and verdict-recording path at `ForkChoiceStoreRun`.

PR #82 introduced the successful-write theorem for `recordPayloadInclusionListSatisfaction`. This follow-up characterizes the two collector fork definitions used by the recorder and proves the recorder’s complete run equation across every outcome.

The following principal equations are tagged:

- `getInclusionListCommittee_run_eq`
  - `@[characterizes EthCLSpecs.Heze.getInclusionListCommittee]`
- `getInclusionListTransactions_run_eq`
  - `@[characterizes EthCLSpecs.Heze.getInclusionListTransactions]`
- `recordPayloadInclusionListSatisfaction_run`
  - `@[characterizes EthCLSpecs.Heze.recordPayloadInclusionListSatisfaction]`

## What changed

Adds `Proofs/Heze/GetInclusionListTransactions.lean` with:

- the complete committee `.run` equation;
- the exact empty-committee arithmetic error;
- the compositional `.run` equation for `getInclusionListTransactions`;
- `FirstReachableMissingTimeliness` over the actual `FcMap.fold` entries;
- the exact `.missingKey ilRoot` result for the first reachable missing timeliness key; and
- unchanged propagation of committee-construction and collection errors.

Extends `Proofs/Heze/RecordPayloadInclusionListSatisfaction.lean` with:

- the principal whole-run recorder equation;
- the slot-zero error corollary;
- arbitrary collector-error propagation;
- the empty-committee consequence; and
- the missing-timeliness consequence.

The existing successful-path theorem from PR #82 remains available. It states the exact returned `FcMap.insert` update with an arbitrary collector-produced `postRunnerStore`.

The proof ledger and implementation notes are updated, and proof coverage is regenerated to reflect all three characterized fork definitions.

### Slot zero

```lean
.error (.transition (.arithmetic
  "record_payload_inclusion_list_satisfaction: Slot(state.slot - 1)"))
```

This occurs before transaction collection.

### Empty inclusion-list committee

```lean
.error (.transition (.arithmetic
  "get_inclusion_list_committee: indices[i % len(indices)] on an empty committee"))
```

The committee error propagates through transaction collection and the recorder unchanged.

### Missing timeliness key

For the first reachable non-equivocator entry whose timeliness lookup is missing:

```lean
.error (.missingKey ilRoot)
```

The collector error propagates through the recorder unchanged.

### Arbitrary collector error

Any other `StoreTransitionError` returned by `getInclusionListTransactions` is returned unchanged.

### Successful collection

If collection returns `.ok (ilTxs, postRunnerStore)`, the recorder returns:

```lean
.ok (
  { store with
    payloadInclusionListSatisfaction :=
      FcMap.insert store.payloadInclusionListSatisfaction root
        (isInclusionListSatisfied payload ilTxs) },
  postRunnerStore)
```

At `StateT σ (Except StoreTransitionError)`, an error result contains no runner state. On success, the recorder returns exactly the runner state produced by transaction collection.

## Characterization policy

`getInclusionListCommittee` and `getInclusionListTransactions` are fork definitions with exact `.run` equations at the selected runner, so their principal equations are tagged `@[characterizes]`.

`getInclusionListTransactions_run_eq` is compositional: it exposes the exact sequencing of committee construction and `collectInclusionListTransactions`.

`collectInclusionListTransactions` is a plain `def`, not a fork definition, and remains untagged. Supporting branch and propagation theorems also remain untagged.

## Test plan

All checks were rerun after rebasing onto current `upstream/main` and after applying the final characterization annotations.

- [x] `lake build`
- [x] `lake build EthCLSpecs`
- [x] `just lint`
- [x] `just check-citations`
- [x] `just proof-coverage-update`
- [x] `just proof-coverage-check`
- [x] `just ethcl-test`
- [x] `just test`
- [x] `git diff --check`
- [x] `#check` and `#check @...` on every public declaration
- [x] `#print axioms` on every public theorem
- [x] Forbidden-feature and scratch-file scan

## Risk and trust footprint

No new trust mechanisms are introduced. The public theorems use only the existing kernel axioms `propext`, `Classical.choice`, and `Quot.sound`, or subsets thereof.

The PR adds no Mathlib dependency, `sorry`, `native_decide`, `bv_decide`, `unsafe`, `partial def`, or new `@[extern]`.

## Notes for reviewers

- The recorder proof target is `ForkChoiceStoreRun (Store map)`.
- The committee equation is polymorphic over the runner-state type because committee construction neither reads nor changes that state.
- The successful recorder result proves the exact structural `FcMap.insert`.
- It does not claim that a later generic lookup returns the inserted value, because the generic `FcMap` interface provides no insert/lookup law.
- The separate same-root payload/verdict pairing and lookup obligation remains tracked under `onExecutionPayloadEnvelope`.
- This is the complete-characterization follow-up to PR #82; PR #82 remains the origin of the successful-path theorem.

## LICENSE (LGPL version 3)

- [x] I have read and accepted LICENSE (LGPL version 3), NOTICE, and CLA in the root of this project.


packages/EthCLSpecs/EthCLSpecs/Proofs/Gloas/Run.lean
```diff
@@ -56,43 +58,33 @@ abbrev GloasRun [Preset] [HasherTag] : Type → Type :=
...
-`PUnit`-valued loop bodies without an instantiation dance. -/
-
-/-- `.run` of a bind: run the first action, and on success run the continuation from the
-value and state it produced. The `Except` bind on the right short-circuits a reject. -/
+/-- See `ForkChoiceStoreRun.run_bind`. -/
 theorem GloasRun.run_bind {σ ε α β : Type} (x : StateT σ (Except ε) α)
     (f : α → StateT σ (Except ε) β) (s : σ) :
-    (x >>= f).run s = (x.run s) >>= fun p => (f p.1).run p.2 := rfl
+    (x >>= f).run s = (x.run s) >>= fun p => (f p.1).run p.2 :=
+  ForkChoiceStoreRun.run_bind x f s
```
leolara: Whys is better to change it to do the proof based on ForkChoiceStoreRun?

packages/EthCLSpecs/EthCLSpecs/Proofs/Heze/GetInclusionListTransactions.lean
```diff
@@ -0,0 +1,331 @@
...
+open EthCLLib.Spec (HasherTag MapKind FcMap throwArithmetic htr StoreTransitionError arrayUnion)
+open EthCLSpecs.Heze (Preset Store State Root Slot ValidatorIndex InclusionList Transaction
+  InclusionListStore getInclusionListCommittee getInclusionListTransactions
+  collectInclusionListTransactions getBeaconCommittee getCommitteeCountPerSlot
+  computeEpochAtSlot cyclicSample)
+open EthCLSpecs.Heze.Const (inclusionListCommitteeSize)
+
+/-- `.run` of `throwArithmetic` at the fork-choice store runner. Closes by `rfl`
+without unfolding `liftErr`. -/
+private theorem throwArithmetic_run {σ α : Type} (descr : String) (s : σ) :
```
leolara: this theorem seems repeated twice

leolara:
@Mouzayan  We should add something like this to the proof ledger:

```md
 | `collectInclusionListTransactions` | `Heze/ForkChoice.lean:135-152` | The collected set: an
  equivocator's entry contributes nothing; at `onlyTimely = true` an entry contributes exactly when
  its stored timeliness is `true`; at `onlyTimely = false` every non-equivocator entry contributes;
  the result is `arrayUnion` of the contributions, so it holds no duplicate.
  `getInclusionListTransactions_run_eq` pins the arguments this is called at and leaves the returned
  array unconstrained | proposed |  |
```

and remove the `plain def` mentions of it, I think we need a characterisation for this one as well.

packages/EthCLSpecs/EthCLSpecs/Proofs/Gloas/Run.lean
Mouzayan: You are right that this equation is not specific to ForkChoiceStoreRun. The intention was to share the definitional `StateT σ (Except ε)` bind equation, but placing it under ForkChoiceStoreRun creates a misleading dependency from the Gloas proof module.

I have prepared a fork neutral `EthCLSpecs.Proofs.Run` module in the follow up characterization work PR #102. 
I will bring that refactor into this PR so both fork-specific modules can use the shared equation.

packages/EthCLSpecs/EthCLSpecs/Proofs/Heze/GetInclusionListTransactions.lean
Mouzayan: Yes, thank you. It is duplicated verbatim as a private helper in `GetInclusionListTransactions.lean` and `RecordPayloadInclusionListSatisfaction.lean`. I will move it into the shared `StoreRun` module use the shared theorem at the two call sites.

Mouzayan:
> @Mouzayan We should add something like this to the proof ledger:
> 
> ```
>  | `collectInclusionListTransactions` | `Heze/ForkChoice.lean:135-152` | The collected set: an
>   equivocator's entry contributes nothing; at `onlyTimely = true` an entry contributes exactly when
>   its stored timeliness is `true`; at `onlyTimely = false` every non-equivocator entry contributes;
>   the result is `arrayUnion` of the contributions, so it holds no duplicate.
>   `getInclusionListTransactions_run_eq` pins the arguments this is called at and leaves the returned
>   array unconstrained | proposed |  |
> ```
> 
> and remove the `plain def` mentions of it, I think we need a characterisation for this one as well.

Understood. 
The current proof covers the missing timeliness error, but not which transactions are returned on success or how duplicates are removed. 
I’ll add that missing work to the proof ledger and remove the wording that suggests no characterization is needed.

